PR #4, opened 2017-03-30
Move encoder initialisation into function

andresailer:
BEGINRELEASENOTES
- RecoMCTruthLinker: move encoder initialisation from constructor to function body to avoid accessing encoding_string too early

ENDRELEASENOTES


andresailer:
I have merged this now, so it will be in the nightly build for tomorrow

PR #9, opened 2017-05-31
read parameters from dd4hep, fix iterator in tau candidate quality loop

webermat:
BEGINRELEASENOTES
- read magnetic field from dd4hep in taufinder particle preparator
- remove leftover condition for former printout of the taufinder, which prevents the desired increment of the iterator, and thus deletes wrong taucandidates when filling the reconstructed tau's 
ENDRELEASENOTES

### DIFF
--- a/Analysis/TauFinder/src/PrepareRECParticles.cc
+++ b/Analysis/TauFinder/src/PrepareRECParticles.cc
@@ -25,8 +25,8 @@ using namespace std;
 #include "IMPL/LCFlagImpl.h" 
 #include "UTIL/LCRelationNavigator.h"
 
-#include <gear/GEAR.h>
-#include <gear/BField.h>
+#include "DD4hep/LCDD.h"
+#include "DD4hep/DD4hepUnits.h"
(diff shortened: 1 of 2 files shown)


Analysis/TauFinder/src/PrepareRECParticles.cc
```diff
@@ -111,7 +111,11 @@ void PrepareRECParticles::init()
   
   // usually a good idea to
   printParameters() ;
-  _bField = Global::GEAR->getBField().at( gear::Vector3D( 0., 0., 0.) ).z() ;
+  DD4hep::Geometry::LCDD& lcdd = DD4hep::Geometry::LCDD::getInstance();
```
andresailer: Please use the utility function from marlinutil geometryutils to get the bfield

webermat:
suggested change using the GeometryUtil functionality for calling the BField

andresailer:
The other thing was my fault. I had mistakenly switched the if and printout lines :(

PR #10, opened 2017-06-15
 Set default from 'ON' to 'OFF' to build MarlinReco fortran sources.

shaojunlu:
BEGINRELEASENOTES
- Set default from 'ON' to 'OFF' to build MarlinReco fortran sources.
- It could be switch on in the release-ilcsoft.cfg if you want.
- https://github.com/iLCSoft/iLCInstall/blob/master/releases/HEAD/release-ilcsoft.cfg#L116

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ SET( ENV{CFLAGS} "-w -ansi -pedantic $ENV{CFLAGS}" ) # -w suppress all warning m
 
 ####################################################################################################
 
-OPTION( MARLINRECO_FORTRAN "Set to ON to build MarlinReco fortran sources" ON )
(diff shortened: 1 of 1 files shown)


petricm:
If I remember correctly @gaede once said that we could actually drop the fortran code from MarlinReco, since is it not used and not even compatible with currant workflow. If this is the case, this might be the right moment to remove it?

shaojunlu:
Dear Marko, and Frank, 

I have included Jenny and Robert for any discussion about this issue. 

Cheers 
Shaojun

gaede:
@petricm MarlinReco needs some general cleaning up, where obsolete stuff gets removed.  When we do this, we can also remove the Fortran stuff.

PR #24, opened 2017-09-26
Fixed the issue of exception

libo929:
BEGINRELEASENOTES
- improve SDHCALDigi/src/SimDigital.cc
     - Fixed the exception of unknown name y
      - add optional parameter HCALCellSize to overwrite the one from dd4hep

ENDRELEASENOTES


gaede:
I don't think that a hard coded cell size is a good idea - at least you should make this an optional processor parameter that can be used to overwrite the value from the DD4hep model if really needed.  We will soon create reconstruction models in lcgeo for the sdhcal that will then have the correct settings.

PR #27, opened 2017-11-06
Temporary fix BitField64 use in SDHCAL digitizer

rete:
BEGINRELEASENOTES
- Change use of dd4hep::BitField64 class by lcio::BitField64 class
- Adapt the different methods to adapt for this change

ENDRELEASENOTES


rete:
Do not merge without approval of SDHCAL team...

gaede:
The original issue in dd4hep::BitField64 is fixed here: https://github.com/AIDASoft/DD4hep/pull/256. So in principle this is no longer needed.

PR #33, opened 2017-12-01
Dress electrons in IsolatedLeptonFinder

andb0t:
This addition to the IsolatedLeptonFinderProcessor dresses electrons with close-by photons. It is currently under development and this PR serves as baseline for discussion.

BEGINRELEASENOTES

  - This package is an extension to the IsolatedLeptonFinderProcessor. The default functionality is untouched.
  - Optionally, it dresses leptons with close-by particles. By default it dresses electrons and muons with photons.
  - The algorithm starts with the highest energy lepton and adds all photons (and, optionally, electrons) in a cone of a given size around to it. As the original, it creates a collection with the dressed leptons and another collections with all remaining particles, except the ones that were dressed into the leptons.
  - All compiler warnings are fixed

ENDRELEASENOTES

### DIFF
--- a/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
+++ b/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
@@ -161,9 +161,9 @@ class IsolatedLeptonFinderProcessor : public Processor {
 
 
 		/** If set to true, uses lepton dressing */
-		float _dressPhotonCosConeAngle = 0;
-		float _mergeLeptonCosConeAngle = 0;
 		std::string _whichLeptons = "UNDRESSED";
+		float _dressPhotonConeAngle = 0;
+		float _mergeLeptonConeAngle = 0;
 		std::vector<int> _dressedPFOs {};
 
 		/** If set to true, uses Pandora particle IDs */
--- a/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
+++ b/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
@@ -161,9 +161,9 @@ class IsolatedLeptonFinderProcessor : public Processor {
 
 
 		/** If set to true, uses lepton dressing */
-		bool _useLeptonDressing = false;
 		float _dressPhotonCosConeAngle = 0;
 		float _mergeLeptonCosConeAngle = 0;
+		std::string _whichLeptons = "UNDRESSED";
 		std::vector<int> _dressedPFOs {};
 
 		/** If set to true, uses Pandora particle IDs */
--- a/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
+++ b/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
@@ -31,11 +31,11 @@ class IsolatedLeptonFinderProcessor : public Processor {
 		virtual Processor*  newProcessor() { return new IsolatedLeptonFinderProcessor ; }
 
 		IsolatedLeptonFinderProcessor() ;
+		IsolatedLeptonFinderProcessor(const IsolatedLeptonFinderProcessor &);
+		IsolatedLeptonFinderProcessor & operator = (const IsolatedLeptonFinderProcessor &);
 
 		virtual void init() ;
-		virtual void processRunHeader( LCRunHeader* run ) ;
-		virtual void processEvent( LCEvent * evt ) ; 
-		virtual void check( LCEvent * evt ) ; 
+		virtual void processEvent( LCEvent * evt ) ;
 		virtual void end() ;
 
 	protected:
@@ -55,10 +55,10 @@ class IsolatedLeptonFinderProcessor : public Processor {
 		bool IsLepton( ReconstructedParticle* pfo ) ;
 
 		/** Returns true if it passes impact parameter cuts */
-		bool PassesImpactParameterCuts( ReconstructedParticle* pfo ) ; 
+		bool PassesImpactParameterCuts( ReconstructedParticle* pfo ) ;
 
 		/** Returns true if it passes impact parameter significance cuts */
-		bool PassesImpactParameterSignificanceCuts( ReconstructedParticle* pfo ) ; 
+		bool PassesImpactParameterSignificanceCuts( ReconstructedParticle* pfo ) ;
 
 		/** Calculates the cone energy */
 		float getConeEnergy( ReconstructedParticle* pfo ) ;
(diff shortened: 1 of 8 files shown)


andb0t:
The basic functionality is implemented now. It does not alter the default behavior but simply adds new output collections or replaces the default ones if the corresponding flag is set. I would be happy about comments!

gaede:
I have asked @tomohikosan for a review, as he is one of the original authors of this processors, so we should wait for him to comment.
From my side:
- your release notes are a bit sparse
    - could you add some documentation on the dressing algorithm used ?
    - you could also add that you fixed all warnings 
- you seem to have added a new feature for optionally formatting the code 
    - this should be documented somewhere
    - which standard did you use ?
    - does it work on all platforms ?

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -242,31 +254,63 @@ IsolatedLeptonFinderProcessor::IsolatedLeptonFinderProcessor()
...
+	// Output PFOs removed isolated leptons
 	LCCollectionVec* otPFOsRemovedIsoLepCol = new LCCollectionVec( LCIO::RECONSTRUCTEDPARTICLE ) ;
 	otPFOsRemovedIsoLepCol->setSubset(true) ;
 
 	// Output PFOs of isolated leptons
 	LCCollectionVec* otIsoLepCol = new LCCollectionVec( LCIO::RECONSTRUCTEDPARTICLE );
 	otIsoLepCol->setSubset(true);
 
+	// Output PFOs removed dressed isolated leptons
+	LCCollectionVec* otPFOsRemovedDressedIsoLepCol = new LCCollectionVec( LCIO::RECONSTRUCTEDPARTICLE ) ;
```
andresailer: As it is optional which collections are stored you need to cleanup. Easiest with unique_ptr

```c++
atuo* otPFOsRemovedDressedIsoLepCol = std::unique_ptr<LCCollectionVec>(new LCCollectionVec( LCIO::RECONSTRUCTEDPARTICLE );
```
and then later
```c++
evt->addCollection(otPFOsRemovedDressedIsoLepCol.release(), "collectionName");
```
If the collection is not added it gets cleaned up automatically


PS: Would be nice to have c++14 and make_unique

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -280,38 +324,212 @@ void IsolatedLeptonFinderProcessor::processEvent( LCEvent * evt ) {
...
-		<< "   in run:  " << evt->getRunNumber() 
-		<< std::endl ;
 
-	// Add PFOs to new collection
-	evt->addCollection( otPFOsRemovedIsoLepCol, _outputPFOsRemovedIsoLepCollection.c_str() );
-	evt->addCollection( otIsoLepCol, _outputIsoLepCollection.c_str() );
-}
+	// Dressed leptons
+	// order by energy
+	for (unsigned int i = 1; i < goodLeptonIndices.size(); ++i)
```
andresailer: use `std::sort` to do the ordering

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -280,38 +324,212 @@ void IsolatedLeptonFinderProcessor::processEvent( LCEvent * evt ) {
...
+	{
+		if (isMoreEnergetic(i, i-1)) {
+			std::swap(goodLeptonIndices.at(i), goodLeptonIndices.at(i-1));
+			i = 1;
+		}
+	}
+	// dress them
+	for (unsigned int i = 0; i < goodLeptonIndices.size(); ++i)
+	{
+		ReconstructedParticle* pfo_tmp = dynamic_cast<ReconstructedParticle*>( _pfoCol->getElementAt(goodLeptonIndices.at(i) ));
```
andresailer: if you don't check for `nullptr` you can use `static_cast`

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
```diff
@@ -31,103 +33,145 @@ class IsolatedLeptonFinderProcessor : public Processor {
 		virtual Processor*  newProcessor() { return new IsolatedLeptonFinderProcessor ; }
 
 		IsolatedLeptonFinderProcessor() ;
+		IsolatedLeptonFinderProcessor(const IsolatedLeptonFinderProcessor &);
```
andresailer: Copy and assignment operator are never used so ` ... = delete;` both of them

andresailer:
@gaede the clang-format file is the one we use in ConformalTracking, which IIRC was similar to what was being done in DD4hep (unless Andreas applied modifications). it works on Mac OS and linux if clang-format is in the path, but it was not applied here. Andreas kept things consistent with the existing code base of the IsolatedLeptonFinder.

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
andb0t: This is to fix this warning at compile time:
```shell
warning: 'class IsolatedLeptonFinderProcessor' has pointer data members
warning: but does not override 'IsolatedLeptonFinderProcessor(const IsolatedLeptonFinderProcessor&)
warning: or 'operator=(const IsolatedLeptonFinderProcessor&)'
```

As it is now, the IsolatedLeptonFinder has not a single compiler warning. Do you know another way to circumvent that? If the warning is senseless we should disable it in the compiler settings.

andb0t:
@gaede and @andresailer, it is exactly the clang-format file Andre mentions. At the moment the `make format` target only formats the IsolatedLeptonFinder. I planned to have a separate PR later with the code format applied and, optionally, format the entire code base.

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
andresailer: I mean you can do
```c++
IsolatedLeptonFinderProcessor(const IsolatedLeptonFinderProcessor &) = delete;
```
to tell the compiler do remove this function

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
```diff
@@ -19,6 +20,7 @@
 
 #include <EVENT/ReconstructedParticle.h>
 #include <EVENT/MCParticle.h>
+#include "IMPL/ReconstructedParticleImpl.h"
 #include <UTIL/LCRelationNavigator.h>
 
 using namespace lcio ;
```
petricm: Please remove `using namespace` from global context of header files

gaede:
Following Tomohiko's comment: could you please rename your processor, e.g. to `IsolatedLeptonFinderNew` and leave the old one untouched. Then people can run the two in parallel and convince themselves that both are equivalent if used w/o dressing the leptons. 
Later we can then replace the old one with the new.

tomohikosan:
@gaede I would maybe call it `DressedLeptonFinder` and not replace the old processor. This is to (1) ensure reproducibility of existing code not just today but also in the future as more development could be made to the new processor, and (2) retain the choice for future extensions to be based on the simpler old processor.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
andb0t: Hi Andre,
your suggestion yields the following error:
```
error: unable to deduce ‘auto*’ from ‘std::unique_ptr<IMPL::LCCollectionVec>((std::__cxx11::basic_string<char>(EVENT::LCIO::RECONSTRUCTEDPARTICLE, std::allocator<char>()), (operator new(256ul), (<statement>, ((IMPL::LCCollectionVec*)<anonymous>)))))’
  auto* otPFOsRemovedDressedIsoLepCol = std::unique_ptr<LCCollectionVec>(new LCCollectionVec( LCIO::RECONSTRUCTEDPARTICLE ) );

```

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
andb0t: Thanks, I was looking for a solution with my lambda function and sort but gave up at some point. The `not` is not necessary, it needs a greater comparison.

petricm:
@tomohikosan so you are basically arguing that `IsolatedLeptonFinder` is frozen and it should never be touched again?

I am not particularly in favor of making this a new processor:
 - @andb0t fixed all the warnings in`IsolatedLeptonFinder`, if he makes a new processor the bad state of the code remains in the `IsolatedLeptonFinder`. Why not accept the fixes and improvements made to the code? Following such approach we will never start reducing the warnings in MarlinReco and just have a lot of duplicate code.
 - if somebody want's the old processor, we have a version control system, you can pull it out anytime you want, or you can use an older tag.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
andb0t: Done

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
andb0t: That removes half the warning, this one remains:
```
warning: ‘class IsolatedLeptonFinderProcessor’ has pointer data members [-Weffc++]
 class IsolatedLeptonFinderProcessor : public Processor {
warning:   but does not override ‘operator=(const IsolatedLeptonFinderProcessor&)’ [-Weffc++]
```

tomohikosan:
@petricm Ok, I was not aware the old processor was producing warnings. The solution is simple and the fixes to reduce them should also be made to the old processor. This is physics code and its *behavior* should not change in any major way once it has been used in validated physics studies, in this case for the DBD. I believe that duplication is a price to way if we are to ensure reproducibility of our results. My primary objection is that if we accept these changes now (which will indeed most likely reproduce the results today), I see no guarantee that someone following a similar line of development in the future will be just as careful. If you have a solution to this, I'd be happy to see it replace the old code.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
andb0t: Done! The `auto*` -> `auto` was the solution

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
andb0t: So this is done as well!

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
andb0t: Done!

tomohikosan:
I have taken a closer look with the assumption that this will replace the old processor.
1. I'm not particularly happy with the way the protected methods have been hacked with the `bool omitDressed` flag. If we were to further extend this processor with feature X, we will need to add another `omitX` flag into the different methods -- this is clearly not the ideal way to extend the processor! A more elegant solution is to convert `_pfoCol` into something like `vector<ReconstructedParticle*> _workingList` where you select only the particles you wish to keep and then apply the different algorithms serially.
2. There is no need for producing output collections for both versions -- again, if we were to add another feature X, this will mean it will produce 2x(# of flags) output collections. Instead, this can be simply achieved by running different instances of Marlin processors with different flags as needed, as the computational burden here is very low.
3. Documentation should be added.

andb0t:
> I have taken a closer look with the assumption that this will replace the old processor.

Great, I think this is what the review is for. I'll adress all your comments until everybody is happy. That way we keep the codebase clean and structured.

>     I'm not particularly happy with the way the protected methods have been hacked with the bool omitDressed flag. If we were to further extend this processor with feature X, we will need to add another omitX flag into the different methods -- this is clearly not the ideal way to extend the processor! A more elegant solution is to convert _pfoCol into something like vector<ReconstructedParticle*> _workingList where you select only the particles you wish to keep and then apply the different algorithms serially.

That's a good suggestion! I can surely find a nice way to make it less "hacked".

> There is no need for producing output collections for both versions -- again, if we were to add another feature X, this will mean it will produce 2x(# of flags) output collections. Instead, this can be simply achieved by running different instances of Marlin processors with different flags as needed, as the computational burden here is very low.

That's a very small change in the code and we can discuss it.

> Documentation should be added.

Surely, I'll add documentation, once we have finished the general review.

gaede:
@tomohikosan I fully agree with all your comments! 
 If these would be addressed and we could keep only one version of the IsolatedLeptonFinder that everybody is happy with, I believe we'd have found the best solution. In the long run it is always better to have one well maintained algorithm than having two slightly different and diverging versions to basically do the same thing...

gaede:
@andb0t are you still working on the suggested changes ?

andb0t:
Yes, I'm back from holidays now and will start working ASAP.

andb0t:
Hello all,
I have updated the files to address @tomohikosan 's comments:

>   1.  I'm not particularly happy with the way the protected methods have been hacked with the bool  omitDressed flag. If we were to further extend this processor with feature X, we will need to add another omitX flag into the different methods -- this is clearly not the ideal way to extend the processor! A more elegant solution is to convert _pfoCol into something like vector<ReconstructedParticle*> _workingList where you select only the particles you wish to keep and then apply the different algorithms serially.
   
I have replaced the flag by the vector as you suggested. This vector is updated when the leptons are dressed such that the merged objects do not enter the cone algorithms.

> 2. There is no need for producing output collections for both versions -- again, if we were to add another feature X, this will mean it will produce 2x(# of flags) output collections. Instead, this can be simply achieved by running different instances of Marlin processors with different flags as needed, as the computational burden here is very low.

The additional collections have been removed. The standard behavior is the original behavior and when the flag 'DRESSED' is passed, the output collections are filled with dressed objects instead.

> 3.  Documentation should be added.

I updated the Readme and the `steer.xml` example file.

Additionally, I have added a useful makefile target to apply the iLCSoft style. The software here is not written in this style and as soon as we are happy with the content, I will apply the correct formatting.

Please review and comment!

gaede:
@tomohikosan could you have a look, if this addresses your comments ?

tomohikosan:
Thanks for the changes. This is one step in the direction I had in mind. However, I still have a concern about the fact that the undressed and dressed versions of the algorithms are run in two separate loops. I would suggest the following procedure:
1. Create a boolean flag to turn dressing on/off (and please not a string...)
2. Prepare a working list of particles (as you have done)
3. If the dressing flag is set, loop through the working list and call dressLepton()
4. Loop through the updated working list and apply the nominal selections
5. Create output collections based on the surviving working list
If you do it this way, I believe steps 1,2,4,5 can be common code. This will be important in case one wants to add new features in the future. Instead of having to modify both pieces of the code (dressed/undressed), one can simply extend 4.

andb0t:
>     Create a boolean flag to turn dressing on/off (and please not a string...)

Done

>    [...] If you do it this way, I believe steps 1,2,4,5 can be common code. This will be important in case one wants to add new features in the future. Instead of having to modify both pieces of the code (dressed/undressed), one can simply extend 4.

Done. There is now one loop to filter out the lepton indices, one to process the leptons (being either dressed or left untouched, depending on the 'useDressedLeptons' boolean flag), and one to process the remaining non-lepton particles.

PR #34, opened 2017-12-10
Initialize all members and remove few unused parameters

petricm:
BEGINRELEASENOTES
- Remove all warnings of type `should be initialized in the member initialization list [-Weffc++]`
- Remove all warnings of type `unused parameter ‘run’` for `processRunHeader( LCRunHeader*  /*run*/)`
- Remove all warnings of type `unused parameter ‘evt’` for `check( LCEvent *  /*evt*/ )`

ENDRELEASENOTES
This reduces the log output from the compile for 2500 lines, the resulting Travis-CI output should be much more comprehensible now.


petricm:
I will remove the changes regarding the IsolatedLaptonFinder as not to cause a conflict.

Regarding the rest of the warnings:

1. I think there are several warnings that might indicate something like
```
../Analysis/IsolatedLeptonTagging/src/Utilities.cc:98:109: warning: unused parameter 'iHiggs' [-Wunused-parameter]

 Int_t getOriginalSerial(ReconstructedParticle *recPart, LCCollection *colMCTL, LCCollection *colMCP, Bool_t iHiggs) {

                                                                                                             ^~~~~~

../Analysis/IsolatedLeptonTagging/src/Utilities.cc: In function 'void isolep::dumpJetParticles(EVENT::ReconstructedParticle*, EVENT::LCCollection*, EVENT::LCCollection*)':

../Analysis/IsolatedLeptonTagging/src/Utilities.cc:796:67: warning: unused parameter 'colMC' [-Wunused-parameter]

   void dumpJetParticles(ReconstructedParticle *jet, LCCollection *colMC, LCCollection *colMCTL) {

                                                                   ^~~~~
```
which I left on purpose. As there are several such places, where a functions takes an argument and I don't know if the author intended to use this parameter and this is an actual bug, of if this is just sloppy. The same holds for warnings from `Wunused-but-set-variable`.

2. I will sometime later (another PR) try to address warnings of the type
  - `has pointer data members`
  - `warning:   but does not override`
  - `warning:   or 'operator=`

3. I think the rest of the warnings might indicate actual problems in the code

petricm:
All done

PR #44, opened 2018-03-15
update dEdx processor: geomtry issue for low momentum; new dx strategies

ueinhaus:
BEGINRELEASENOTES
- Compute_dEdXProcessor:
  - geometry issue: gear to DD4hep unit adaption, fixed low momentum problem
  - added dx calculation strategies
  - added various processor options, default are all old version
  - added documentation

ENDRELEASENOTES

### DIFF
--- a/Analysis/PIDTools/src/Compute_dEdxProcessor.cc
+++ b/Analysis/PIDTools/src/Compute_dEdxProcessor.cc
@@ -8,6 +8,7 @@
 
 #include <marlin/Global.h>
 #include <marlin/Processor.h>
+#include <marlin/AIDAProcessor.h>
 
 #include <lcio.h>
 #include <EVENT/LCCollection.h>
@@ -17,9 +18,19 @@
 
 #include <DD4hep/Detector.h>
 #include <DDRec/DetectorData.h>
+#include <DD4hep/DD4hepUnits.h>
 
 #include <UTIL/BitField64.h>
 #include "UTIL/LCTrackerConf.h"
+
+// MarlinUtil
+#include "LCGeometryTypes.h"
+#include "SimpleHelix.h"
+#include "HelixClass.h"
+
+#include "TCanvas.h"
+#include "TImage.h"
+#include "TStyle.h"
 
 #include "Compute_dEdxProcessor.hh"
 
@@ -31,11 +42,32 @@ Compute_dEdxProcessor::Compute_dEdxProcessor()
   // Processor description
   _description = "dE/dx calculation using truncation method" ;
   
+  registerInputCollection(LCIO::TRACK,
+			  "LDCTrackCollection",
+			  "LDC track collection name",
+			  _LDCTrackCollection,
+			  std::string("MarlinTrkTracks"));
+
+  registerProcessorParameter( "Write_dEdx",
+			      "If set, the calculated dEdx value will be attached to its corresponding track (default: true).",
+			      _writedEdx,
+			      bool(true));
+
   registerProcessorParameter( "EnergyLossErrorTPC",
 			      "Fractional error of dEdx in the TPC",
 			      _energyLossErrorTPC,
 			      float(0.054));
   
+  registerProcessorParameter( "LowerTruncationFraction",
+			      "Lower truncation fraction, default: 0.08 (8%)",
+			      _lowerTrunFrac,
+			      float(0.08));
+
+  registerProcessorParameter( "UpperTruncationFraction",
+			      "Upper truncation fraction, default: 0.3 (30%)",
+			      _upperTrunFrac,
+			      float(0.3));
+
   registerProcessorParameter( "isSmearing",
 			      "Flag for dEdx Smearing",
 			      _isSmearing,
@@ -46,9 +78,7 @@ Compute_dEdxProcessor::Compute_dEdxProcessor()
 			      _smearingFactor,
 			      float(0.035));
 
-  std::vector<float> apar;
-  apar.push_back(0.635762);
-  apar.push_back(-0.0573237);
+  std::vector<float> apar = {0.635762, -0.0573237};
   registerProcessorParameter( "AngularCorrectionParameters",
 			      "parameters for angular correction",
 			      _acorrpar,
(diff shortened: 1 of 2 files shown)


rete:
@gaede @andresailer TravisCI does not seem to start. Nothing has been created on TravisCI page and it's been more than a day now. Any idea on how to (re-)trigger it ?

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
```diff
@@ -104,136 +177,305 @@ void Compute_dEdxProcessor::processEvent( LCEvent * evt ) {
...
+//  for(int iH=0;iH<itpc;iH++){
+//    for(int jH=iH+1;jH<itpc;jH++){
+//      if(depe[iH]<depe[jH]){
+//        tmpe=depe[iH];
+//        depe[iH]=depe[jH];
+//        depe[jH]=tmpe;
+//      }
+//    }
+//  }
+  std::sort(dEdxVec1.begin(),dEdxVec1.end());
```
rete: Thanks for that ! ;-)

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
```diff
@@ -104,136 +177,305 @@ void Compute_dEdxProcessor::processEvent( LCEvent * evt ) {
...
+//    //TrackerHit * hit = hitExt->getTrackerHit();
+//    if(float(depe[iH])>mine && float(depe[iH])<maxe){   //need to check here
+//      //if(float(depe[iH])<maxe){   //need to check here
+//      dedx+=depe[iH];
+//      nTruncate++;
+//      //biH=iH;
+//    }
+//  }
+
+  for (int idE=dEdxVec1.size()*_lowerTrunFrac; idE<dEdxVec1.size()*(1-_upperTrunFrac); idE++) {dedx1 += dEdxVec1[idE]; nTruncate1++;}
```
rete: Can you use your option _dxStrategy to make a choice on the dEdX calculation and avoid to compute it 3 times. If I choose an option, I don't want to spend time on calculation for the 2 others options I don't want.

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
```diff
@@ -104,136 +177,305 @@ void Compute_dEdxProcessor::processEvent( LCEvent * evt ) {
...
+  case  3: normdedx = normdedx3; nTruncate = nTruncate3; totway = totover;   break;
+  default: normdedx = normdedx1; nTruncate = nTruncate1; totway = totlength;
+  }
 
-  float *ret=new float[2];
-  ret[0]=normdedx;
-  ret[1]=normdedx*std::pow(0.001*totlength, -0.34)*std::pow(nTruncate, -0.45);   //todo: what is gas pressure in TPC?
+  std::pair<double,double> ret;
+  ret.first =normdedx;
+  ret.second=normdedx*std::pow(0.001*totway, -0.34)*std::pow(nTruncate, -0.45);   //todo: what is gas pressure in TPC?
```
rete: Can you translate these -0.34 and -0.45 parameters into something we can put in the processor parameter as well ?

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
```diff
@@ -104,136 +177,305 @@ void Compute_dEdxProcessor::processEvent( LCEvent * evt ) {
...
 
-  if(ret[0]!=ret[0]){
-    ret[0]=0.0;
-    ret[1]=0.0;
+  if(ret.first!=ret.first){
```
rete: If you want to check for NAN as the next lines suggest, use `std::isnan(ret.first)` instead.

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
```diff
@@ -250,7 +492,7 @@ float Compute_dEdxProcessor::getNormalization(double dedx, float hit, double trk
 }
 
 //create smearing factor
-float Compute_dEdxProcessor::getSmearing(float dEdx){
+double Compute_dEdxProcessor::getSmearing(double dEdx){
   double z=sqrt( -2.0*std::log(dist(*engine)) ) * std::sin( 2.0*3.141592*dist(*engine) );
```
rete: Can you also turn this ugly 3.141592 into M_PI from the `cmath` header ?

andresailer:
@rete close and open triggered travis
@ueinhaus please add the processor your changes are for to the release notes

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
```diff
@@ -59,28 +89,71 @@ Compute_dEdxProcessor::Compute_dEdxProcessor()
...
+    const int nBinsX=1000, nBinsY=300;
+    double minBinX=-1, maxBinX=2;  // log10(min/max momentum / GeV)
+    double minBinY= 0, maxBinY=1e-6;
+    double histbinsX[nBinsX+1], histbinsY[nBinsY+1];
+    for (int i=0; i<nBinsX+1; i++) histbinsX[i] = pow( 10, minBinX + (maxBinX-minBinX)*i/(float)(nBinsX) );
+    for (int i=0; i<nBinsY+1; i++) histbinsY[i] = minBinY + (maxBinY-minBinY)*i/(float)(nBinsY);
+
+    // create the histograms
+    AIDAProcessor::tree(this);
+    _BBHist_Strategy1 = new TH2D("BBHist_Strategy1","Bethe-Bloch curve for dx strategy 1: hit-to-hit distance",nBinsX,histbinsX,nBinsY,histbinsY);
```
rete: `AIDAProcessor::tree(this)` does not assume you are using a ROOT implementation. You will get memory leak in this case.
If you want to use the AIDA processor, then use the [AIDA histogram factory](https://github.com/iLCSoft/Marlin/blob/master/source/include/marlin/AIDAProcessor.h#L84) : 
```cpp
auto factory = AIDAProcessor::histogramFactory(this);
auto histo = factory->createHistogram2D("BBHist_Strategy1",  "Bethe-Bloch curve for dx strategy 1: hit-to-hit distance", nBinsX, minBinX, maxBinX, nBinsY, minBinY, maxBinY);
// example of filling later in the code
histo->fill(mom, dedx);
```
Your second solution is to write your histogram in a separate root file. Then you need an additional processor parameter for the file name. In this case you can also write your TImage in this TFile instead of a separate png file.

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: done

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: done

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: I have to ask Masakazu, I don't know what these numbers are.

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: :D

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: done

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
gaede: @rete technically you are of course right - but RAIDA is the only AIDA implementation that we ever used. So I think it is fine to instantiate ROOT histos after a call to `AIDAProcessor::tree(this);` Then every ROOT object goes to the directory with the processor name in the main (RAIDA)ROOT file ...

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: done

Analysis/PIDTools/src/Compute_dEdxProcessor.cc
ueinhaus: Ok, so If I can leave it like that, then I'm done with the changes and would attempt a new pull request - permission granted?

andresailer: You don't have to make a new pull request, just push to your `improveDeDx` branch, it will update the pull request.

rete:
I will update ILDConfig accordingly with the default new parameters. Which strategy should I specify ? According to [your talk](https://agenda.linearcollider.org/event/7906/contributions/41121/attachments/32884/50039/2018_03_21_SwAna_Coll_Meeting.pdf), strategy 1 and 3 give similar results. I would be rather conservative and keep strategy 1 to run the same algorithm as before. What do you think ?

ueinhaus:
<processor name="MyCompute_dEdxProcessor" type="Compute_dEdxProcessor"> 
 
<parameter name="LDCTrackCollection" type="string"> MarlinTrkTracks </parameter> 
 
<parameter name="EnergyLossErrorTPC" type="float"> 0.047 </parameter> 
<parameter name="Write_dEdx" type="bool"> true </parameter> 
<parameter name="LowerTruncationFraction" type="float"> 0.08 </parameter> 
<parameter name="UpperTruncationFraction" type="float"> 0.3 </parameter> 
<parameter name="isSmearing" type="bool"> true </parameter> 
<parameter name="smearingFactor" type="float"> 0.035 </parameter> 
<parameter name="AngularCorrectionParameters" type="std::vector<float>"> 0.635762 -0.0573237 </parameter> 
<parameter name="NumberofHitsCorrectionParameters" type="float"> 1.468 </parameter> 
<parameter name="dxStrategy" type="int"> 1 </parameter> 
<parameter name="StrategyCompHist" type="bool"> false </parameter> 
<parameter name="StrategyCompHistWeight" type="bool"> false </parameter> 
<parameter name="StrategyCompHistFiles" type="std::string"> dEdx_Histo_Strategy </parameter> 
</processor> 


So the strategy stays 1 at the moment. 
And energy loss should be 4.7%, and smearing 3.5%. 

MfG Uli

rete:
OK nice thank you !
Our current configuration also says : 
```xml
<!--dEdx smearing factor  (0.035 for l5, 0.044 for s5)-->
<parameter name="smearingFactor" type="float"> 0.035 </parameter>
```
Is the number for the small model (0.044) also correct ?

PR #49, opened 2018-04-10
WIP: Add tof

gaede:
BEGINRELEASENOTES
- add new package TimeOfFlight
     - use TOFEstimators processor to compute TOF parameters
     - will be added as PID object to the ReconstructedParticles

ENDRELEASENOTES

### DIFF
--- a/TimeOfFlight/src/TOFEstimators.cc
+++ b/TimeOfFlight/src/TOFEstimators.cc
@@ -34,7 +34,7 @@
 #include <gsl/gsl_randist.h>
 
 //---- ROOT -----
-#include "TH2F.h" 
+#include "TH2F.h"
 
 using namespace lcio ;
 using namespace marlin ;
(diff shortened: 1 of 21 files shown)


TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -0,0 +1,435 @@
...
+
+
+      
+      // =======  use only Ecal hits  (requires the CalorimeterHitType to be set in the digitizer )
+      //          with time information ( > 1 ps) and layer <= max layer
+
+      int maxLayerNum = _maxLayerNum ;
+      std::function<bool(CalorimeterHit*)> selectHits =  [maxLayerNum](CalorimeterHit* h){
+
+	return ( isEcal( h )            &&
```
rete: Why do you use only ecal hits ? What about neutral hadrons ?

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -0,0 +1,435 @@
...
+
+    colPFO = nullptr ;
+  }
+    
+  if( colPFO != nullptr ){
+    
+
+
+    PIDHandler pidh( colPFO );
+    int algoID = pidh.addAlgorithm( name()  , _TOFNames);
```
rete: Maybe we should add a processor parameter for this. If it is not provided, we can jump to a fallback with the processor name. What do you think ?

TimeOfFlight/src/TOFEstimators.cc
gaede: At this point it is a feature that we restrict ourselves to the Ecal only. Combining hits from Ecal and Hcal with different time resolutions would probably require a more involved algorithm.
We can leave this for a future version.

TimeOfFlight/src/TOFEstimators.cc
gaede: Actually, I think this is more elegant as the Marlin framework requires that different processors of the same type have different names. If you had an additional parameter you'd have to change both consistently in the steering file. What would be the advantage ?

PR #67, opened 2019-06-03
SimpleFcalDigi fix (cellID1)

rete:
BEGINRELEASENOTES
- SimpleFcalDigi processor:
    - Missing cellid 1 flag in fcal output collection, causing issues in the BeamCal reconstruction

ENDRELEASENOTES


rete:
@gaede FYI

gaede:
Thanks.  Two questions:
- is there no digitzer for FCal in the new package from @danieljeans ?
- do we need similar bug fixes in there ?

PR #82, opened 2020-08-31
Introduce some newlines to check if travis CI kicks in

tmadlener:
Dummy PR to see if travis has a general issue with this repository or if it is related to specific users.


rete:
Looks like it's working dude...

PR #83, opened 2020-11-26
RealisticCaloDigi: new ROC integration

### LINKED ISSUE #41: Time realism of Calo Digi
The current implementation in the RealisticCaloDigi is estimated the time of a digitized hit as the time of the MC contribution with the earliest time:
https://github.com/iLCSoft/MarlinReco/blob/master/CaloDigi/Realistic/src/RealisticCaloDigi.cc#L291

It was pointed out that a more realistic time calculation should be estimated as follows:
- Sort MC contributions by increasing time
- Loop over contributions and accumulate deposit energy until you reach the digi threshold
- Estimate the time of the digitized hit as the time of the contribution that first reach the threshold

Does this have to be changed, or are we OK with the current (approximate) implementation ?

rete:
BEGINRELEASENOTES
- RealisticCaloDigi: Added new option for energy/charge and time integration
  - deal with slow/fast shaper of ROC chips and time estimate
  - new processor parameters:
     - **integrationMethod**: "Standard" for old implementation (default) or "ROC" emulating the behavior of ROC chip.
     - **fastShaper**: fast shaper time, unit in ns. Only for ROC method
     - **slowShaper**: slow shaper time, unit in ns. Only for ROC method
     - **timingResolution**: optional time resolution gaussian smearing to apply, unit . Only apply is > 0. Default is 0 (no smearing)

ENDRELEASENOTES

Fixes #41 

Parameter values for ROC implementation:
- AHCAL
   - **fastShaper** = 15 ns
   - **slowShaper** =  50 ns
   - **timingResolution** = 700 ps
- SiWECal
   - **fastShaper** = 90 ns (first estimate)
   - **slowShaper** =  180 ns (first estimate)
   - **timingResolution** = 700 ps (first guess from AHCAL value and the fact that the TDC is similar. Not a real estimate)

TODO:
- [x] Add details in release notes
- [x] Validate changes in the code for previous version (identical distributions)
- [x] Determine whether the ROC implementation works for the SiWECal electronics
- [x] Determine parameters for all technologies
- [x] Make plots:
    - rec time distribution
    - rec energy distribution
- [x] Document timing treatment in a Readme.md

### DIFF
--- a/CaloDigi/Realistic/doc/Readme_ROCTiming.md
+++ b/CaloDigi/Realistic/doc/Readme_ROCTiming.md
@@ -1,0 +1,57 @@
+## New timing treatment in the `RealisticCaloDigi` processor (03.2021)
+
+### Possible integration methods: `Standard` VS `ROC`
+
+The RealisticCaloDigi processor includes two different timing treatment methods, also influencing the energy integration.
+
+- The `Standard` integration sums up the energy of Monte Carlo contributions (MCCs) of *SimCalorimeterHits* within a certain time window (`t_min` < t < `t_max`). The time stamp of the reconstructed hit is set to the earliest MCC time of the simulated hit and the energy is set to the sum of the MCC energies within the time window.
+- The `ROC` (ReadOut Chip) integration reflects a bit more how the electronics designed by the Omega group behaves. These chips include a fast shaper `t_fast` and a slow shaper `t_slow` value which are use in this algorithm to perform the energy integration and hit time estimate. Starting from the earliest MCC, the algorithm checks if the accumulated energy within `t_fast` passes the energy threshold. If not, it goes to the next MCC and performs the same check. If the threshold is reached, the hit time is set to the current MCC time and the energy is set to the accumulated MCC energy within `t_slow` from the current MCC. An absolute time cut is finally applied. If the hit time is in a certain time window (`t_min` < t < `t_max`), it is accepted, else discarded. If the hit is accepted, a random gaussian smearing is (optionally) applied to the hit time.
+
+### Processor parameters
+
+Here are listed the processor parameters acting on timing.
+
+General parameters:
+
+- `timingWindowMin`: The minimum hit time to accept a hit (t > `t_min`). Float, unit in `ns`, default is -10.
+- `timingWindowMax`: The maximum hit time to accept a hit (t < `t_max`). Float, unit in `ns`, default is 100.
+- `integrationMethod`: The energy integration and timing estimate method for digitization. Options are "Standard" (default) and "ROC".
+
+Parameters for the "Standard" method **only**:
+
+- `timingCorrectForPropagation`: Hit time correction `t_corr` for propagation computed as radial distance divided by `c`. Hit time becomes t - t_corr. Int, default is 0 (false).
+
+Parameters for the "ROC" method **only**:
+
+- `fastShaper`: The fast shaper integration time `t_fast`. Must be set in the steering file if the ROC method is chosen. Float, unit in `ns`, default is 0.
+- `slowShaper`: The slow shaper integration time `t_slow`. Must be set in the steering file if the ROC method is chosen. Float, unit in `ns`, default is 0.
+- `timingResolution`: Apply a random gaussian smearing `t_res` on the final hit time. The parameter value corresponds to the width of the gaussian function. Only applied if greater than 0. Float, unit in `ns`, default is 0 (no time smearing).
+
+The default behavior of the digitizer is then:
+
+- `integrationMethod`: "Standard"
+- `timingWindowMin`: -10 ns
+- `timingWindowMax`: 100 ns
+- `timingCorrectForPropagation`: false
+
+
+### Additional remarks on the ROC implementation
+
+#### Pull request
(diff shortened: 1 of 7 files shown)


rete:
@gaede @ebrianne @dudarboh @JennyListDESY @danieljeans 
I would like to start a discussion on the implementation of a realistic time estimation in the ILD digitizers. Here is a PR, flagged as WIP. I will update the release notes to give more details soon. The original implementation of the ROC integration was taken from there: https://stash.desy.de/projects/CALICE/repos/calice_sim/browse/digitization/digiAhc2/src/Ahc2ROCThresholdProcessor.cc
Note that with the default settings at the moment, the old behaviour is kept for backward compatibility.

danieljeans:
this must be done in very close collaboration with the hardware groups, and should explicitly be signed off by them.
I think it's probably a non-trivial thing to get right.

ebrianne:
I have been already discussing with @rete about it yesterday and explaining why and how it is done for the AHCAL in TB. 
I agree that this would need to be validated by each hardware groups before the switch. I think we did studies some time ago on the impact of the threshold on timing which is minimal but this will I guess depend on the time resolution, especially at the ps level and may need to be looked into again. 
However, one has to be careful, we noticed not so long ago that the threshold has quite some impact on the energy below 1 MIP (at the 0.5 MIP threshold border). This is due to how we do our digitization workflow. You can contact Olin about this as he did the study and see if this would impact the ILDDigi workflow.

rete:
@airqui ping

airqui:
Hi @rete , thanks for the ping. I am indeed interested on this, specially on the test beam context, but I always manage to push it down in the pipeline of things to do...
I am up for discussions on this and to contribute to the si-ECAL part...

rete:
The new implementation (ROC) works for the AHCAL electronics, as most of the code was taken from the AHCAL TB digitizer.
@airqui Would you mind going through the implementation and see whether it would work for the ECal electronics too?
Also we would need to determine a few parameter values as quoted in the top-level PR message. I assumed from a discussion with @ebrianne that the HCal values are correct.

airqui:
Hi @rete,
sorry for the late reply. In the SiECAL, the efforts on the simulation/tb analysis have been somehow limited in the latest times since the personpower has been focused on the hardware activities. I plan to start looking at the digitization this January and hopefully we could conduct some comparison with old data during the first months of the year. This means that:

1. In January/February I will carefully check how we can adapt the AHCAL TB digitizer to our purposes (in principle our digitize is a simplified version of theirs)
2. The optimization of the needed parameters will take some time and I am not in position of making an educated guess of how long will it take..

Maybe it is not needed that we fully optimize the TB digitizer before work in the ILD digitizer. 

Question for @ebrianne : I imagine that the best way to start (thinking on test beam data) is to start with your AHCAL TB digitizer, right? Better than going directly to the ILD digitizer... am I correct?

ebrianne:
I think this is the easiest way to start things. Technically I guess not much needs to be adapted but this depends on your framework I guess.

rete:
@ebrianne @airqui @danieljeans  I have checked that the old integration (standard) is backward compatible. It looks all fine. 

In the new ROC implementation there seems to be a feature that I have not taken into account. As we now integrate using slow/fast shapers time parameters, there is not cut on the absolute hit time anymore. You can see it below for 30 GeV K0L, hcal hits only.

Full plot:

Plot_MyRecHCalLCTuple_Time

Zoom around the 100 ns cut value:

Plot_MyRecHCalLCTuple_TimeZoom

I have thus added in the latest commits (https://github.com/iLCSoft/MarlinReco/pull/83/commits/918489f81ae184639d3ec742a3dfd198163f2b5a) a similar time window cut using the same processor parameters for the ROC implementation. But now I'm not sure about the meaning of this cut in terms of electronics behavior. These late hits have to be cut away in any case but does this cut, **in the way it is implemented now**, make sense for the electronics. Isn't there a more meaningful way of applying this cut ?

As we've been discussing internally at DESY also, I will merge this pull request now, as time is getting short for me and as we will be running the standard integration in any case in the production and for some time. 

Give me your comments ! :-)

gaede:
Hi Remi, as far as I understand it, this looks good - provided that the default for the late time cut is set to infinity or very large times. The calo experts should comment, if they think differently ....

ebrianne:
Hi @rete, this looks good to me. I agree with @gaede that the default time cuts should be put as high as possible. One may be interested in the future to look at late time hit in hadron showers.

CaloDigi/Realistic/doc/Readme_ROCTiming.md
```diff
@@ -0,0 +1,57 @@
...
+#### Current ROC parameter values for known technologies (03.2021)
+
+- AHCal:
+  - `fastShaper`: 15 ns
+  - `slowShaper`: 50 ns
+  - `timingResolution`: 700 ps
+- SiWEcal:
+  - `fastShaper`: 90 ns. First estimate
+  - `slowShaper`: 180 ns. First estimate
+  - `timingResolution`: 700 ps. First guess from AHCAL value, as the TDC is similar. Not a real estimate
```
ebrianne, line 52: Can we get a number of the SiW ECAL prototype? I would think it is in the same ballpark or even better.

rete: @airqui ping

tmadlener (approved):

PR #86, opened 2021-01-25
Migrate CI to github actions

tmadlener:
BEGINRELEASENOTES
- Migrate CI from travis to github actions

ENDRELEASENOTES

Following similar migration for LCIO (iLCSoft/LCIO#125).

- [x] add github actions
- [x] coverity scan(?) (keeping track of this with #86)
- [x] trigger rebuild of iLCSoft (`after_success` in travis) (also keeping this in #86)
- [x] remove travis workflows

The main thing missing from the current github actions is the coverity scan and the trigger for the rebuild of iLCSoft. However I couldn't find how those were solved for LCIO.

@andresailer @rete I can't seem to request a review, so I have to ping you like this.


tmadlener:
Results for the github actions until they are also triggered on this repository: https://github.com/tmadlener/MarlinReco/actions

rete:
I guess coverity scan has not been running properly since ages... Not a problem to be removed, except if there are complains :-)

.github/workflows/linux.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        COMPILER: [gcc, llvm]
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cvmfs-contrib/github-action-cvmfs@v2
+    - name: Start containter
```
rete, line 15: Containter -> Container

.github/workflows/linux.yml
```diff
@@ -0,0 +1,20 @@
...
+      run: |
+        docker run -it --name CI_container -v /home/runner/work/MarlinReco/MarlinReco:/Package -v /cvmfs:/cvmfs:shared -e COMPILER=${{ matrix.COMPILER }} -d ghcr.io/aidasoft/centos7:latest /bin/bash
```
rete, line 17: Is there a builtin variable for the package name here? It would be nice to use it because then we could re-use the same script for all iLCSoft packages

tmadlener: Indeed, I just [learned](https://docs.github.com/en/actions/reference/environment-variables) you can use `GITHUB_WORKSPACE` for this.

petricm:
> I guess coverity scan has not been running properly since ages... Not a problem to be removed, except if there are complains :-)

There was a problem with coverity and I have been waiting for a new release since ages, which was finally produces 2 days ago `2020.09` (https://community.synopsys.com/s/coverity-scan-status).

tmadlener:
So should we try to set this up again with github actions somehow? There doesn't seem to be a simple packaged solution yet, but it also doesn't seem to be impossible to do so: https://github.com/ruby/actions-coverity-scan/blob/master/.github/workflows/coverity-scan.yml

Not sure how hard it would be in this case to get this running again.

tmadlener:
I think I have a potentially working setup: https://github.com/tmadlener/MarlinReco/blob/enable-coverity/.github/workflows/coverity-scan.yml
Currently this fails because the token is not accessible from within the workflow. It should be possible to add this token to the secrets (maybe even globally for iLCSoft), and then this workflow might work. Then we would need to schedule it properly and also add proper uploading of the results.

petricm:
> I think I have a potentially working setup: https://github.com/tmadlener/MarlinReco/blob/enable-coverity/.github/workflows/coverity-scan.yml
> Currently this fails because the token is not accessible from within the workflow. It should be possible to add this token to the secrets (maybe even globally for iLCSoft), and then this workflow might work. Then we would need to schedule it properly and also add proper uploading of the results.

The tokens are per-project

tmadlener:
Ah right, then it would need to be added at the organization level if we want coverity scans: https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization

rete:
I personally don't have any access to these tokens. Can somebody take care of adding them in Github?

tmadlener:
Alternatively, we can re-add the coverity scans in a separate PR, since this one is essentially finished, once the travis CI yaml file is removed.

README.md
```diff
@@ -1,5 +1,6 @@
 # MarlinReco
 [![Build Status](https://travis-ci.org/iLCSoft/MarlinReco.svg?branch=master)](https://travis-ci.org/iLCSoft/MarlinReco)
```
rete, line 2: Can you also remove this badge then? :-)

tmadlener: Done. :)

rete (approved):

PR #88, opened 2021-02-08
Addition of new algorithms to TOFEstimator processor

dudarboh:
BEGINRELEASENOTES

- TOFEstimators processor: adding new algorithms to calculate the ToF, track length and momentum.
  - New output parameters:
    - TOFClosest --- based on the closest ECAL hit
    - TOFFastest --- based on the fastest ECAL hit
    - TOFCylFit --- extrapolated time from the ECAL hits within a cylinder of a shower core
    - TOFClosestFit --- extrapolated time from the ECAL hits closest to the linear continuation of the track inside the ECAL
    - FlightLength --- the helix track length based on the track parameters from the TrackState at the ECAL
    - MomAtCal --- the momentum based on the track parameters from the TrackState at the ECAL
  - New steering parameters:
    - ProcessorVersion --- changes output between the idr version (before this patch) and the dev (after this patch). Default: idr
    - CylRadius --- the radius within which to select hits for the TOFCylFit method. Default: 5 mm

ENDRELEASENOTES

### Important notes
Currently new algorithms work only with simple cases of PFO, meaning: **only one cluster** and **only one track** associated with PFO. Other PFOs are ignored.

Flight length calculation are done using the assumption of less than 2*pi change in the phi angle. Meaning that the **flight length** gives the **wrong** results for the tracks with **more than one helix curls**.

Thus, the new algorithms are still optimized to work in the barrel part only. **Performance** of the TOF estimators in the **endcap** is currently **unknown**.

### What's new
We added new TOF estimators that show a better performance in terms of less bias in the measurement of pions, kaons and protons masses.

### Steering parameter
New steering parameter `ProcessorVersion` has been added with the possible values: `"idr"` (default) and `"dev"`.
The previous behavior is set by default for the backwards compatibility.

### Output
In the new version the output parameters are modified to:
`{"TOFClosest", "TOFFastest", "TOFCylFit", "TOFClosestFit", "FlightLength", "MomAtCalo"}`

- TOF* --- are TimeOfFlight estimated by different algorithms (see bellow).
- FlightLength --- is a track length of a particle from IP (assuming (0, 0, 0)) to ECAL surface
- MomAtCalo --- Track momentum at ECAL surface.

Using these parameters the mass of a particle can be calculated.

### TOF estimators
The four estimators are available in the new version:
- TOFClosest --- time of the ECAL hit closest to the extrapolated track position at the ECAL surface with correction for distance between impact point and center of ECAL cell assuming speed of light propagation.
- TOFFastest --- time of the ECAL hit with the smallest time with correction for distance between impact point and center of ECAL cell assuming speed of light propagation.
- TOFCylFit --- time estimated with a linear fit of the ECAL hits times in the 5mm cylinder with the main axis parallel to track direction at the ECAL surface vs distance to the impact point of each hit. Then TOF is the bias parameter of a linear fit.
- TOFClosestFit --- time estimated with a linear fit of the ECAL hits times which are closest to the track line estimated at the ECAL surface (one closest hit per layer) vs distance to the impact point of each hit. Then TOF is the bias parameter of a linear fit.

### DIFF
--- a/TimeOfFlight/include/TOFEstimators.h
+++ b/TimeOfFlight/include/TOFEstimators.h
@@ -3,22 +3,36 @@
 
 #include "marlin/Processor.h"
 #include "lcio.h"
+
+//std stuff
+#include <iostream>
 #include <string>
 #include <vector>
+#include <random>
+#include <functional>
+#include <memory>
+using std::string, std::vector, std::endl;
+
 #include <gsl/gsl_rng.h>
+
+#include <EVENT/Track.h>
+#include <EVENT/Cluster.h>
+
+#include "DDRec/Vector3D.h"
+using dd4hep::rec::Vector3D;
 
 using namespace lcio ;
 using namespace marlin ;
 
-
-class TH2F ;
+#include "TH2F.h"
+// class TH2F ;
 
 /** Compute estimators for the time of flight from the CalorimeterHits in Clusters.
  *  The estimators are stored in a PID object with the name of the processor.
  *  The following parameters are stored:
  *   TOFFirstHit         -  first hit ( closest to calo entry point)
  *   TOFClosestHits      -  closest hit in every layer (<lMax)
- *   TOFClosestHitsError -  error of above 
+ *   TOFClosestHitsError -  error of above
  *   TOFFlightLength     -  trajectory length to reference point
  *   TOFLastTrkHit       -  (unsmeared) time of last tracker hit (in SET)
  *   TOFLastTrkHitFlightLength -  trajectory length to last trk hit
@@ -29,69 +43,65 @@ class TH2F ;
  *  TimeResolution  - assumed single hit resolution in ps
  *
  *  Only Ecal hits are considered.
- *  For charged particles the TrackState at the calorimeter is used as reference point and for the 
- *  extrapolation into the calorimeter. For neutral particles the position of the first hit (the one closest to the IP) 
+ *  For charged particles the TrackState at the calorimeter is used as reference point and for the
+ *  extrapolation into the calorimeter. For neutral particles the position of the first hit (the one closest to the IP)
  *  is used and a straight flight path from the IP is taken for the extrapolation.
  *  The hit time is corrected for the flight time from the reference assuming speed of light.
- *  
+ *
  *  See ../scripts/tofestimators.xml for example steering file.
  *
  * @author F.Gaede, DESY, April 2018
+ * @author B.Dudar, DESY, February 2021
  */
 
 class TOFEstimators : public Processor {
-  
- public:
-  
-  virtual Processor*  newProcessor() { return new TOFEstimators ; }
-  
-  TOFEstimators(const TOFEstimators&) = delete;
-  TOFEstimators& operator=(const TOFEstimators&) = delete;
+    public:
 
-  TOFEstimators() ;
-  
-  /** Called at the begin of the job before anything is read.
-   * Use to initialize the processor, e.g. book histograms.
-   */
-  virtual void init() ;
-  
-  /** Called for every run.
-   */
-  virtual void processRunHeader( LCRunHeader* run ) ;
-  
-  /** Called for every event - the working horse.
-   */
-  virtual void processEvent( LCEvent * evt ) ; 
-  
-  
-  virtual void check( LCEvent * evt ) ; 
-  
-  
-  /** Called after data processing for clean up.
-   */
-  virtual void end() ;
-  
-  
- protected:
+    virtual Processor*  newProcessor() { return new TOFEstimators ; }
 
-  /** Input collection name with ReconstructedParticles (PFOs).
-   */
-  std::string _colNamePFO{};
-  int   _maxLayerNum{} ;
-  float _resolution{} ;
-  
-  std::vector<std::string> _TOFNames {} ; 
+    TOFEstimators(const TOFEstimators&) = delete;
+    TOFEstimators& operator=(const TOFEstimators&) = delete;
 
-  int _nRun{};
-  int _nEvt{};
+    TOFEstimators() ;
 
-  gsl_rng* _rng = nullptr ;
+    /** Called at the begin of the job before anything is read.
+    * Use to initialize the processor, e.g. book histograms.
+    */
+    virtual void init() ;
 
-  std::vector<TH2F*> _h{};
+    /** Called for every event - the working horse.
+    */
+    virtual void processEvent(LCEvent* evt) ;
 
-} ;
+    virtual void check(LCEvent* evt) ;
+
+    /** Called after data processing for clean up.
+    */
+    virtual void end() ;
+
+    Vector3D getMomAtCalo(const Track*);
+    double getFlightLength(const Track*);
+    double getTOFClosest(const Track*, const Cluster*);
+    double getTOFFastest(const Track*, const Cluster*);
+    double getTOFCylFit(const Track*, const Cluster*);
+    double getTOFClosestFit(const Track*, const Cluster*);
+
+    protected:
+
+    /** Input collection name with ReconstructedParticles (PFOs).
+    */
+    string _colNamePFO{};
+
+    string _procVersion{};
+    int _maxLayerNum{};
+    float _resolution{};
+    vector<string> _TOFNames{};
+
+    gsl_rng* _rng = nullptr;
+    vector<TH2F*> _h{};
+
+    std::default_random_engine _generator{};
+    std::normal_distribution<double> _smearing{};
+};
 
 #endif
-
-
-
--- a/TimeOfFlight/include/TOFEstimators.h
+++ b/TimeOfFlight/include/TOFEstimators.h
@@ -11,7 +11,7 @@
 #include <random>
 #include <functional>
 #include <memory>
-using std::string, std::vector, std::endl;
+#include <limits>
 
 #include <gsl/gsl_rng.h>
 
@@ -19,13 +19,11 @@ using std::string, std::vector, std::endl;
 #include <EVENT/Cluster.h>
 
 #include "DDRec/Vector3D.h"
-using dd4hep::rec::Vector3D;
 
 using namespace lcio ;
 using namespace marlin ;
 
 #include "TH2F.h"
-// class TH2F ;
 
 /** Compute estimators for the time of flight from the CalorimeterHits in Clusters.
  *  The estimators are stored in a PID object with the name of the processor.
@@ -79,7 +77,7 @@ class TOFEstimators : public Processor {
     */
     virtual void end() ;
 
-    Vector3D getMomAtCalo(const Track*);
+    dd4hep::rec::Vector3D getMomAtCalo(const Track*);
     double getFlightLength(const Track*);
     double getTOFClosest(const Track*, const Cluster*);
     double getTOFFastest(const Track*, const Cluster*);
@@ -90,18 +88,21 @@ class TOFEstimators : public Processor {
 
     /** Input collection name with ReconstructedParticles (PFOs).
     */
-    string _colNamePFO{};
+    std::string _colNamePFO{};
 
-    string _procVersion{};
+    std::string _procVersion{};
     int _maxLayerNum{};
+    double _cylRadiusCut{};
     float _resolution{};
-    vector<string> _TOFNames{};
+    std::vector<std::string> _TOFNames{};
 
     gsl_rng* _rng = nullptr;
-    vector<TH2F*> _h{};
+    std::vector<TH2F*> _h{};
 
-    std::default_random_engine _generator{};
+    std::mt19937 _generator{};
     std::normal_distribution<double> _smearing{};
+
+    double _bField[3];
 };
 
 #endif
--- a/TimeOfFlight/include/TOFEstimators.h
+++ b/TimeOfFlight/include/TOFEstimators.h
@@ -25,29 +25,81 @@ using namespace marlin ;
 
 #include "TH2F.h"
 
-/** Compute estimators for the time of flight from the CalorimeterHits in Clusters.
- *  The estimators are stored in a PID object with the name of the processor.
- *  The following parameters are stored:
- *   TOFFirstHit         -  first hit ( closest to calo entry point)
- *   TOFClosestHits      -  closest hit in every layer (<lMax)
- *   TOFClosestHitsError -  error of above
- *   TOFFlightLength     -  trajectory length to reference point
- *   TOFLastTrkHit       -  (unsmeared) time of last tracker hit (in SET)
- *   TOFLastTrkHitFlightLength -  trajectory length to last trk hit
- *
- *  processor parameters:
- *  MaxLayerNumber  - restrict the hits to the first MaxLayerNumbers
- *  ReconstructedParticleCollection - input collection
- *  TimeResolution  - assumed single hit resolution in ps
- *
- *  Only Ecal hits are considered.
+/**
+ * @section DESCRIPTION
+ * Compute time of flight (ToF) from the CalorimeterHits in the ECAL Cluster.
+ * The ToFs are stored in a PID object with the name of the processor attached
+ * to the PandoraPFOs.
+ * Steering parameters:
+ *  MaxLayerNumber  - Estimate ToF using only MaxLayerNumber first layers in the ECAL
+ *  ReconstructedParticleCollection - input collection (usually PandoraPFOs)
+ *  TimeResolution  - assumed single Calorimeter hit time resolution in ps
+ *  CylRadius - radius within which hits are accepted for the fit. Relevant only
+                for new TOFCylFit algorithm. Default 5 mm
+ *  ProcessorVersion - controls output of the processor between idr
+ *                (interim design report) and dev (current development) versions
+ * Output for ProcessorVersion = "idr"
+ *  The following parameters are stored with the IDR ProcessorVersion:
+ *   TOFFirstHit         -  ToF of the CalorimeterHit closest to track entry
+ *                          point to the ECAL corrected with the distance
+ *                          between cell center and entry point assuming
+ *                          speed of light propagation
+ *   TOFClosestHits      -  ToF estimated as the average of corrected
+ *                          CalorimeterHit times from hits that are closest to
+ *                          the linear prolongation of the trajectory line
+ *                          of the track in MaxLayerNumber first layers
+ *   TOFClosestHitsError -  error of the TOFClosestHits
+ *   TOFFlightLength     -  Track length of the helix between IP and ECAL entrance
+ *                          assuming IP at (0, 0, 0) and ECAL entry point is
+ *                          obtained from Kalman filter by extrapolation.
+ *                          Omega/lambda track parameters are taken from IP hit
+ *                          and assumed to be constant along the track.
+ *                          Results are reasonable for delta phi < 2pi
+ *                          meaning not more than one curl.
+ *   TOFLastTrkHit       -  (unsmeared) time of last tracker hit (in the SET)
+ *   TOFLastTrkHitFlightLength - track length similar to the TOFFlightLength
+ *                               but to the SET hit
+ * Notes:
  *  For charged particles the TrackState at the calorimeter is used as reference point and for the
  *  extrapolation into the calorimeter. For neutral particles the position of the first hit (the one closest to the IP)
  *  is used and a straight flight path from the IP is taken for the extrapolation.
  *  The hit time is corrected for the flight time from the reference assuming speed of light.
  *
+ * Output for ProcessorVersion = "dev"
+ *  The following parameters are stored with the IDR ProcessorVersion:
+ *   TOFClosest         -  ToF of the CalorimeterHit closest to track entry
+ *                          point to the ECAL corrected with the distance
+ *                          between cell center and entry point assuming
+ *                          speed of light propagation
+ *   TOFFastest         -  ToF of the CalorimeterHit arrived fastest corrected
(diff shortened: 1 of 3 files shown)


rete:
@dudarboh Can you:
- fill the release notes between BEGINRELEASENOTES and ENDRELEASENOTES
- rebase on the ilcsoft master. The CI has been updated

?

dudarboh:
Hi, @rete.

Done. Please check.

rete:
@dudarboh When you open the pull request there is message:
```

BEGINRELEASENOTES
- Thank you for writing the text to appear in the release notes. It will show up
  exactly as it appears between the two bold lines
- ...

ENDRELEASENOTES
```

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
-  streamlog_out( DEBUG ) << " --- check called ! " << std::endl ; 
+      // create some histograms with beta vs momentum for charged particles
 
+      if( isFirstEvent() ){
 
-  // create some histograms with beta vs momentum for charged particles
-  
-  if( isFirstEvent() ){
+        // this creates a directory for this processor ....
+        AIDAProcessor::histogramFactory( this ) ;
```
rete, line 390: memory leak?

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+      }
     }
-  }
-  
 }
 
-void TOFEstimators::end(){ 
 
-  gsl_rng_free( _rng );
+void TOFEstimators::end(){ if( _procVersion == "old") gsl_rng_free( _rng ); }
```
rete, line 464: Formatting please

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+Vector3D TOFEstimators::getMomAtCalo(const Track* track){
+    const TrackState* ts = track->getTrackState(TrackState::AtCalorimeter);
+    double phi = ts->getPhi();
+    double d0 = ts->getD0();
+    double z0 = ts->getZ0();
+    double omega = ts->getOmega();
+    double tanL = ts->getTanLambda();
+
+    const Detector& detector = Detector::getInstance();
+    double bField[3];
```
rete, line 476: The B field could be cached in the init function. The call to `detector.field().magneticField({0., 0., 0.}, bField);` could be expensive for every track

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+
+
+double TOFEstimators::getTOFClosest(const Track* track, const Cluster* cluster){
+    const TrackState* ts = track->getTrackState(TrackState::AtCalorimeter);
+    Vector3D trackAtCaloPos = ts->getReferencePoint();
 
-  streamlog_out(MESSAGE) << "TOFEstimators::end()  " << name() 
-			 << " processed " << _nEvt << " events in " << _nRun << " runs "
-			 << std::endl ;
+    double dToImpactMin = 99999.;
```
rete, line 501: use numeric limits please

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+    double hitTime = 0.;
+    for (auto&& hit : cluster->getCalorimeterHits() ){
```
rete, line 503: not `&&` but `&`

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
 }
 
 
-//*******************************************************************************************************
+double TOFEstimators::getTOFFastest(const Track* track, const Cluster* cluster){
+    const TrackState* ts = track->getTrackState(TrackState::AtCalorimeter);
+    Vector3D trackAtCaloPos = ts->getReferencePoint();
+
+    double dToImpactMin = 0.;
+    double hitTimeMin = 99999.;
```
rete, line 526: same here

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+    for (auto&& hit : cluster->getCalorimeterHits() ){
```
rete, line 527: same here

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+        CHT hitType( hit->getType() );
+        bool isEcal = (hitType.caloID() == CHT::ecal);
+        int layer = hitType.layer();
+        if (!isEcal || layer >= _maxLayerNum) continue;
+
+        Vector3D hitPos = hit->getPosition();
+        double dToLine = (hitPos - trackAtCaloPos).cross(trackAtCaloMom.unit()).r();
+        // take only hits from 5 mm cylinders. 5 mm value was ibtain optimizing on photons
+        if (dToLine > 5.) continue;
```
rete, line 560: Should be moved to a processor parameter maybe (?). Not nice to have something hardcoded

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+    Vector3D trackAtCaloMom = getMomAtCalo(track);
+
+    std::map<int, double> dToImpact;
+    std::map<int, double> hitTime;
+    std::map<int, double> dToLineMin;
+    for(int i=0; i<_maxLayerNum; ++i) dToImpact[i] = 0.;
+    for(int i=0; i<_maxLayerNum; ++i) hitTime[i] = 0.;
+    for(int i=0; i<_maxLayerNum; ++i) dToLineMin[i] = 99999.;
+
+    for (auto&& hit : cluster->getCalorimeterHits() ){
```
rete, line 589: same here

rete (changes requested):

TimeOfFlight/include/TOFEstimators.h
```diff
@@ -3,22 +3,36 @@
...
 #include "lcio.h"
+
+//std stuff
+#include <iostream>
 #include <string>
 #include <vector>
+#include <random>
+#include <functional>
+#include <memory>
+using std::string, std::vector, std::endl;
```
tmadlener, line 14: Can we have these at least inside the class definition? Or even better use them in their full glory in headers and maybe only try to shorten them in .cc files when actually necessary. Using things from the `std` namespace this way can lead to interesting things in such a large scope.

TimeOfFlight/include/TOFEstimators.h
```diff
@@ -3,22 +3,36 @@
...
+
 #include <gsl/gsl_rng.h>
 
+#include <EVENT/Track.h>
+#include <EVENT/Cluster.h>
+
+#include "DDRec/Vector3D.h"
+using dd4hep::rec::Vector3D;
```
tmadlener, line 22: Similar thing here. Potentially even more dangerous, because `Vector3D` is more likely to be a user defined class, so "polluting" the global namespace with this might lead to name clashes.

TimeOfFlight/include/TOFEstimators.h
```diff
@@ -29,69 +43,65 @@ class TH2F ;
...
+
+    string _procVersion{};
+    int _maxLayerNum{};
+    float _resolution{};
+    vector<string> _TOFNames{};
+
+    gsl_rng* _rng = nullptr;
+    vector<TH2F*> _h{};
+
+    std::default_random_engine _generator{};
```
tmadlener, line 103: Using the std random number generator facilities is notoriously hard to do properly (see e.g. [here](https://codingnest.com/generating-random-numbers-using-c-standard-library-the-problems/) or [here](https://www.pcg-random.org/other-rngs.html#id28)), are also not guaranteed to be portable. Also the `std::default_random_engine` is implementation defined, so it might be better to use a dedicated generator here, to at least know which one is used.

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -15,7 +12,20 @@
 #include <UTIL/Operators.h>
 #include <UTIL/PIDHandler.h>
 
-#include "DDRec/Vector3D.h"
+#include "EVENT/LCCollection.h"
+#include "EVENT/ReconstructedParticle.h"
+using EVENT::LCCollection, EVENT::ReconstructedParticle;
+
+#include "DD4hep/Detector.h"
+using dd4hep::Detector;
```
tmadlener, line 20: This is only used once below. I would remove it

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,578 @@
...
+
+Vector3D TOFEstimators::getMomAtCalo(const Track* track){
+    const TrackState* ts = track->getTrackState(TrackState::AtCalorimeter);
+    double phi = ts->getPhi();
+    double d0 = ts->getD0();
+    double z0 = ts->getZ0();
+    double omega = ts->getOmega();
+    double tanL = ts->getTanLambda();
+
+    const Detector& detector = Detector::getInstance();
```
tmadlener, line 475:
```suggestion
    const auto& detector = dd4hep::Detector::getInstance();
```
To "compensate" for the `using dd4hep::Detector` at the top.

TimeOfFlight/src/TOFEstimators.cc
rete: Replace with `tree(this)`

TimeOfFlight/scripts/tofestimators.xml
```diff
@@ -13,90 +13,87 @@
...
-  <parameter name="SkipNEvents" value="0" />  
-  <parameter name="SupressCheck" value="false" />  
-  <parameter name="AllowToModifyEvent" value="false" />  
-  <parameter name="GearXMLFile"></parameter>  
-  <parameter name="Verbosity" options="DEBUG0-4,MESSAGE0-4,WARNING0-4,ERROR0-4,SILENT"> MESSAGE DEBUG  </parameter> 
+  <!-- limit the number of processed records (run+evt): -->
+  <parameter name="MaxRecordNumber" value="5001" />
+  <parameter name="SkipNEvents" value="0" />
+  <parameter name="SupressCheck" value="false" />
+  <parameter name="AllowToModifyEvent" value="true" />
```
tmadlener, line 39: Is this intentional?

tmadlener (commented):
Thanks for addressing the comments @dudarboh . I have just one last comment about one change in the xml steering file, that I have missed before. Is that an intentional change or is it left from testing?

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,595 @@
...
-//---- ROOT -----
-#include "TH2F.h"
-
 using namespace lcio ;
 using namespace marlin ;
 using namespace TOFUtils ;
 
-TOFEstimators aTOFEstimators ;
+// Speed of light in mm/ns = 299.79246
+#define SPEED_OF_LIGHT TMath::C()/1.e6
```
rete, line 51: Not better than before. You make a division everytime you use this constant. Use `CLHEP::c_light` instead with the `CLHEP/Units/PhysicalConstants.h` include please

tmadlener: Compiler can probably figure it out in this case. ;)

Nevertheless, using CLHEP you should be able to do
```suggestion
static constexpr double SPEED_OF_LIGHT = CLHEP::c_light;
```

or replace the `SPEED_OF_LIGHT` constant completely by `CLHEP::c_light`. **Beware that c_light` already is in mm/ns as per CLHEP convention**, but if you want to be really sure that this stays that way, even if CLHEP decides to change its basis of units do
```suggestion
static constexpr double SPEED_OF_LIGHT = CLHEP::c_light * CLHEP::mm / CLHEP::ns;
```
and you have explicitly stated what your speed of light is, even with units.

TimeOfFlight/include/TOFEstimators.h
dudarboh: Hi Thomas,

I checked the links, they are very interesting and I agree. But I am not sure if this is so much of importance here. Also there is no such links for gsl generator, but I assume that is because of lack of checking it, rather than because it is perfect. So I would leave std one for the moment if you don't mind, otherwise switch to gsl is not a problem of course

TimeOfFlight/src/TOFEstimators.cc
dudarboh: Is it in mm/ns? if I look in the [header](https://www-zeuthen.desy.de/geant4/clhep-2.0.4.3/PhysicalConstants_8h-source.html) file its in m/s. Am I wrong?

```
 //
00061 // c   = 299.792458 mm/ns
00062 // c^2 = 898.7404 (mm/ns)^2 
00063 //
00064 static const double c_light   = 2.99792458e+8 * m/s;
00065 static const double c_squared = c_light * c_light;
00066 
```

TimeOfFlight/scripts/tofestimators.xml
dudarboh: It is left from testing, yes... The whole xml file i used is only for testing, and as far as I understand it is not used in the production, so I am not sure in which state to leave it for merge... Is it important at all?

TimeOfFlight/include/TOFEstimators.h
tmadlener: Yes, you are right, in this case it probably doesn't matter too much, but in general I think it is good to be aware of these potential problems. I am not sure about gsl either, but in general it should be OK, it just has a more C-like interface, which makes it a bit awkward to use in some cases.

TimeOfFlight/scripts/tofestimators.xml
tmadlener: In that case it isn't really important for production, you are right.
However, does it work without allowing to modify event contents in the tests? If yes, than I would set it to false, even if it is only for testing. Otherwise it is one more thing that you would need to check if things start to break at some point in the future.

TimeOfFlight/src/TOFEstimators.cc
tmadlener: First the header you are looking at is an older version. The version that comes with v02-02 is 2.4.4.3

There the definition is 
https://gitlab.cern.ch/CLHEP/CLHEP/-/blob/CLHEP_2_3_4_3/Units/Units/PhysicalConstants.h#L56

So, still m / s. But if you then check how m and s are defined you will see, that the base units of CLHEP are indeed mm and ns, and the units are defined accordingly:
https://gitlab.cern.ch/CLHEP/CLHEP/-/blob/CLHEP_2_3_4_3/Units/Units/SystemOfUnits.h#L72
https://gitlab.cern.ch/CLHEP/CLHEP/-/blob/CLHEP_2_3_4_3/Units/Units/SystemOfUnits.h#L140

So, if you need a value in specific units, you can transform the numerical value into the CLHEP system by using their units as well.

TimeOfFlight/scripts/tofestimators.xml
dudarboh: Indeed, writing new slcio files works without true flag. So I change back to false

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,591 @@
...
+        _TOFNames = {"TOFFirstHit", "TOFClosestHits", "TOFClosestHitsError", "TOFFlightLength",  "TOFLastTrkHit" , "TOFLastTrkHitFlightLength"};
+    }
+    else if (_procVersion == "dev"){
+        _smearing.param( std::normal_distribution<double>::param_type(0., _resolution/1000.) );
+        _TOFNames = {"TOFClosest", "TOFFastest", "TOFCylFit", "TOFClosestFit", "FlightLength", "MomAtCalo"};
 
+        const auto& detector = dd4hep::Detector::getInstance();
+        detector.field().magneticField({0., 0., 0.}, _bField);
+    }
+    else { throw std::string("Invalid ProcessorVersion parameter passed!!!\n Viable options are: idr (default), dev"); }
```
tmadlener, line 100:
```suggestion
    else { 
        throw EVENT::Exception(std::string("Invalid ProcessorVersion parameter passed: \'") + _procVersion + std::string("\'\n Viable options are idr (default), dev"); 
    }
```
Sorry, missed this before. The important part here is to throw an exception that Marlin/LCIO can properly handle, and also give the user some context when it crashes. Simply throwing a string here potentially gets past all that and simply gives a crash with a stack-trace which is somewhat harder to decipher.
If you include the passed parameter in the exception message is a minor detail in the end.

tmadlener (changes requested):

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,593 @@
...
+    }
+    else if (_procVersion == "dev"){
+        _smearing.param( std::normal_distribution<double>::param_type(0., _resolution/1000.) );
+        _TOFNames = {"TOFClosest", "TOFFastest", "TOFCylFit", "TOFClosestFit", "FlightLength", "MomAtCalo"};
 
+        const auto& detector = dd4hep::Detector::getInstance();
+        detector.field().magneticField({0., 0., 0.}, _bField);
+    }
+    else { 
+        throw EVENT::Exception(std::string("Invalid ProcessorVersion parameter passed: \'") + _procVersion + std::string("\'\n Viable options are idr (default), dev");
```
tmadlener, line 101:
```suggestion
        throw EVENT::Exception(std::string("Invalid ProcessorVersion parameter passed: \'") + _procVersion + std::string("\'\n Viable options are idr (default), dev")); 
```

tmadlener (changes requested):
I missed a parenthesis...

tmadlener (approved):

rete:
@dudarboh  Last thing to sort out probably is the release notes. Can you just provide a few bullet points for summary?

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -33,442 +43,593 @@
...
+    	const double* mom = pfo->getMomentum() ;
+    	double momentum = sqrt( mom[0] * mom[0] +  mom[1] * mom[1] +  mom[2] * mom[2] ) ;
 
-  }
+    	double length  =  tofParams[ tof_length  ] ;
 
-  // get the PFO collection from the event if it exists
-  LCCollection* colPFO = nullptr ;
-  try{ colPFO = evt->getCollection( _colNamePFO ) ; } catch(lcio::Exception){}
+    	double beta_fh = ( length / tofParams[ tof_firsthit] ) / 299.8 ;
```
rete, line 455: Why to use a truncated 299.8 when have the constant available from CLHEP. Please it change everywhere

rete: I know this is not from your algorithms, but if you want to make comparison studies, better not to be biased from this approximation

rete (changes requested):

TimeOfFlight/src/TOFEstimators.cc
dudarboh: It is 2018 Frank's part of the code and changing that will break backwards compatability

dudarboh: You was faster :)
Ok ,will change

dudarboh:
Release notes are now only brief bullet points.

rete (approved):

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -15,7 +12,20 @@
 #include <UTIL/Operators.h>
 #include <UTIL/PIDHandler.h>
 
-#include "DDRec/Vector3D.h"
+#include "EVENT/LCCollection.h"
+#include "EVENT/ReconstructedParticle.h"
+using EVENT::LCCollection, EVENT::ReconstructedParticle;
```
petricm, line 17: This does not work!
```
error: expected ';' before ',' token
 using EVENT::LCCollection, EVENT::ReconstructedParticle;
```

tmadlener: That is interesting. It compiles successfully in the CI workflows. But maybe that is only true for the limited number of compilers that we use there. Where did you test this?

petricm: one of the pipelines that creates the cvmfs nighly https://gitlab.cern.ch/CLICdp/SoftwareConfigurations/iLCSoft/-/pipelines/2360644

gaede: This valid c++17 code that gcc 6 does not seem to support. Do we still need gcc6 to work here ?

petricm: @andresailer has to decide if he still wants that pipeline. But in my opinion it's easier to fix this and keep the current system running.

andresailer: Are we moving to c++17 for iLCSoft?

gaede: I thought we did a long time ago ? At least we are using it for our iLCSoft release > 15 months or so ...

andresailer: I must have missed that memo 😃 

Probably should move to base on the nightly key4hep installation for the CI, which includes the iLCSoft packages

petricm: So we stop the old nightly CI in [this](https://gitlab.cern.ch/CLICdp/SoftwareConfigurations/iLCSoft/pipelines) project?

andresailer: yes

PR #94, opened 2021-08-20
IsolatedLeptonTagging: Always fill expected output collections if inputs are empty

### LINKED ISSUE #93: TaJetClustering crashes with DataNotAvailableException occasionally
We are currently running the [standard miniDST workflow](https://github.com/iLCSoft/ILDConfig/blob/master/StandardConfig/production/MarlinStdRecoMiniDST.xml) which occasionally crashes in the TaJetClustering module if the input collection is not present. We get the following error message:
```console
Vertex collection (PrimaryVertex) is empty !
Vertex collection (PrimaryVertex) is empty !
***********************************************
A runtime error occured - (uncaught exception):
       lcio::DataNotAvailableException: LCEventImpl::getCollection: collection not in event:PFOsminuse
Marlin will have to be terminated, sorry.
***********************************************
```
The first two lines here originate from the IsolatedLeptonTagging processor: https://github.com/iLCSoft/MarlinReco/blob/442d5cdc3b4c2df7b61bd5f6ac485b1bc7ad0fb5/Analysis/IsolatedLeptonTagging/src/IsolatedLeptonTaggingProcessor.cc#L236

This in turn also means that the IsolatedLeptonTagging processor does not produce an output collection. However, the TaJetClustering expects this collection to be present, and doesn't properly handle the case when it isn't:
https://github.com/iLCSoft/MarlinReco/blob/442d5cdc3b4c2df7b61bd5f6ac485b1bc7ad0fb5/Analysis/TauFinder/src/TaJetClustering.cc#L186-L187

tmadlener:
BEGINRELEASENOTES
- Make the IsolatedLeptonTagging processor always produce the expected output collections, even for empty input collections, i.e. if the inputs are empty:
  - the `OutputPFOsWithoutIsoLepCollection` will simply have the same content as the `InputPandoraPFOsCollection`
  -  the `OutputIsoLeptonsCollection` will be empty. 
- On the other hand actually missing input collections will now no longer be handled as these point to a real problem (e.g. typo in the collection name). Fixes #93

ENDRELEASENOTES

### DIFF
--- a/Analysis/TauFinder/src/TaJetClustering.cc
+++ b/Analysis/TauFinder/src/TaJetClustering.cc
@@ -183,8 +183,7 @@ bool compareEnergy( ReconstructedParticle * const &p1, ReconstructedParticle * c
 void TaJetClustering::processEvent( LCEvent * evt ) { 
 
   // obtain Reconstructed Particles
-  LCCollection* col = evt->getCollection( _pfoCollectionName );
-  if( !col ){ /*cout << "Failed to obtain PFO collection!" << endl;*/ return; }
+  LCCollection* col = evt->getCollection(_pfoCollectionName);
   int nPart = col->getNumberOfElements();
   
   // output collection
@@ -457,12 +456,12 @@ void TaJetClustering::processEvent( LCEvent * evt ) {
     pRemainCol->addElement(p);
   }
   evt->addCollection( pRemainCol, _remainCollectionName );
-  
+
   //-- note: this will not be printed if compiled w/o MARLINDEBUG=1 !
-  streamlog_out(DEBUG) << "   processing event: " << evt->getEventNumber() 
-		       << "   in run:  " << evt->getRunNumber() 
+  streamlog_out(DEBUG) << "   processing event: " << evt->getEventNumber()
+		       << "   in run:  " << evt->getRunNumber()
 		       << std::endl ;
-  
+
   _nEvt ++ ;
 }
 
--- a/Analysis/IsolatedLeptonTagging/include/IsolatedLeptonTaggingProcessor.h
+++ b/Analysis/IsolatedLeptonTagging/include/IsolatedLeptonTaggingProcessor.h
@@ -44,6 +44,11 @@ class IsolatedLeptonTaggingProcessor : public marlin::Processor {
   
  protected:
 
+  /**
+   * Add the expected output collections
+   */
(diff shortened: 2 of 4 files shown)


tmadlener:
Seems to fix the TaJetClustering issue, but then the first LCFIPlusProcessor breaks again, because that doesn't handle a missing collection gracefully.

Analysis/TauFinder/src/TaJetClustering.cc
```diff
@@ -183,8 +184,13 @@ bool compareEnergy( ReconstructedParticle * const &p1, ReconstructedParticle * c
 void TaJetClustering::processEvent( LCEvent * evt ) { 
 
   // obtain Reconstructed Particles
-  LCCollection* col = evt->getCollection( _pfoCollectionName );
-  if( !col ){ /*cout << "Failed to obtain PFO collection!" << endl;*/ return; }
+  LCCollection* col{nullptr};
+  try {
+    col = evt->getCollection(_pfoCollectionName);
+  } catch (lcio::DataNotAvailableException& e) {
+    streamlog_out(WARNING) << "No PFO collection (" << _pfoCollectionName << ") found" << std::endl;
```
gaede, line 191: I would not print this as a WARNING if this is expected in some physics events rather use DEBUG5 or so.
Otherwise the logfiles will be swamped with warnings that are none, really

Analysis/TauFinder/src/TaJetClustering.cc
```diff
@@ -457,12 +463,7 @@ void TaJetClustering::processEvent( LCEvent * evt ) {
     pRemainCol->addElement(p);
   }
   evt->addCollection( pRemainCol, _remainCollectionName );
-  
-  //-- note: this will not be printed if compiled w/o MARLINDEBUG=1 !
-  streamlog_out(DEBUG) << "   processing event: " << evt->getEventNumber()
```
gaede, line 462: ... why did you take this DEBUG printout out - it causes no harm and might be useful in a real debugging ?

tmadlener: Mainly because it looked like a leftover from the example on which the processor is based. Additionally, I think it should be at the beginning of `processEvent` rather than at the end to make it really useful. And I didn't find any mention of `MARLINDEBUG` in iLCInstall or the cmake config of Marlin or MarlinReco, so I thought this will not be used in any case.

PR #95, opened 2021-08-23
CI: add github action workflows

andresailer:
BEGINRELEASENOTES
- CI: build against LCG_99python2 gcc8 and LCG_100 gcc10, clang11

ENDRELEASENOTES


tmadlener (changes requested):
https://github.com/iLCSoft/MarlinReco/blob/master/.github/scripts/compile_and_test.sh

Could also be removed with this, since it is no longer necessary.

andresailer:
Done

tmadlener (approved):

tmadlener:
Thanks for this.

PR #98, opened 2021-08-31
fix units in TPC parameters

### LINKED ISSUE #97: Wrong units of TPC parameters in KinkFinder
## Issue

As spotted by @tmadlener [here](https://github.com/iLCSoft/MarlinUtil/issues/22#issuecomment-907192113) KinkFinder uses TPC parameters without specifying specific units.

https://github.com/iLCSoft/MarlinReco/blob/39201c6fbfc257a30e4bb86c7e2eacca908a6729/Tracking/KinkFinder/src/KinkFinder.cc#L198-L201

Program is probably expects values to be in  `dd4hep::mm`, however they are in `dd4hep::cm`.
I checked them with `cout` to confirm that they are indeed in `cm`:
```
[ VERBOSE "MyKinkFinder"] _tpcInnerR: 37.21
[ VERBOSE "MyKinkFinder"] _tpcOuterR: 169.21
[ VERBOSE "MyKinkFinder"] _tpcZmax: 222.5
[ VERBOSE "MyKinkFinder"] _tpcMaxRow: 220
```

It is definitely bug [here1](https://github.com/iLCSoft/MarlinReco/blob/39201c6fbfc257a30e4bb86c7e2eacca908a6729/Tracking/KinkFinder/src/KinkFinder.cc#L757-L783) and [here2](https://github.com/iLCSoft/MarlinReco/blob/39201c6fbfc257a30e4bb86c7e2eacca908a6729/Tracking/KinkFinder/src/KinkFinder.cc#L792-L799) while counting number of TPC hits. As `TrackerHit::getPosition()` returns coordinates in mm.

it is potentially bug [here](https://github.com/iLCSoft/MarlinReco/blob/39201c6fbfc257a30e4bb86c7e2eacca908a6729/Tracking/KinkFinder/src/KinkFinder.cc#L640-L652) as well, as `rkink` gets its value from `HelixClass::getPointInZ()` which probably returns point in `mm` while comparing to values in `cm`

## Relevance

Affects master

gaede:
BEGINRELEASENOTES
- apply the correct units for TPC parameters from DD4hep
     - fixes #97
-  apply correct units for VXD and SIT layers
     -   need cross check of efficiency for kink finding 
     -  (how could this have worked before ?)

ENDRELEASENOTES

### DIFF
--- a/Tracking/KinkFinder/src/KinkFinder.cc
+++ b/Tracking/KinkFinder/src/KinkFinder.cc
@@ -195,9 +195,9 @@ void KinkFinder::init() {
   dd4hep::DetElement tpcDE = theDet.detector("TPC") ;
   const dd4hep::rec::FixedPadSizeTPCData* tpc = tpcDE.extension<dd4hep::rec::FixedPadSizeTPCData>() ;
 
-  _tpcInnerR = tpc->rMinReadout; 
-  _tpcOuterR = tpc->rMaxReadout;
-  _tpcZmax   = tpc->driftLength;
+  _tpcInnerR = tpc->rMinReadout/dd4hep::mm;
+  _tpcOuterR = tpc->rMaxReadout/dd4hep::mm;
+  _tpcZmax   = tpc->driftLength/dd4hep::mm;
   _tpcMaxRow = tpc->maxRow;
 
 
--- a/Tracking/KinkFinder/src/KinkFinder.cc
+++ b/Tracking/KinkFinder/src/KinkFinder.cc
@@ -206,7 +206,7 @@ void KinkFinder::init() {
   _nLayersVTX=vtx->layers.size(); 
 
   for(int iL=0;iL<_nLayersVTX;iL++){
-    _rVTX.push_back( vtx->layers[iL].distanceSensitive );
+    _rVTX.push_back( vtx->layers[iL].distanceSensitive/dd4hep::mm );
   }
 
   dd4hep::DetElement sitDE = theDet.detector("SIT");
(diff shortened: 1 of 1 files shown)


gaede:
Of course, thanks, @dudarboh.

tmadlener:
> - (how could this have worked before ?)

I think we were somewhat lucky because we have built DD4hep such that it does **not** use Geant4 units. In that case `dd4hep::mm == 1.0`. The definition for the version we are using:
https://github.com/AIDASoft/DD4hep/blob/v01-11-02/DDParsers/include/Evaluator/DD4hepUnits.h#L53-L57

(For newer dd4hep versions this can be controlled via the `DD4HEP_USE_GEANT4_UNITS`)

gaede:
But  https://github.com/iLCSoft/MarlinReco/issues/97 suggests that we are using the wrong units (dd4hep::mm=0.1) ?

tmadlener:
Ah right, I missed that. But that would imply that we compile dd4hep with `HAVE_GEANT4_UNITS`? I couldn't find any obvious mention of that in iLCInstall or the CMake config of dd4hep, but it is entirely possible I missed it.

In any case this just proves how important it is to use the correct units ;)

gaede:
> Ah right, I missed that. But that would imply that we compile dd4hep with `HAVE_GEANT4_UNITS`?

No, the opposite. We have `dd4hep::mm=0.1` and thus need the `1/dd4hep::mm` in the code where units are needed...

 >I couldn't find any obvious mention of that in iLCInstall or the CMake config of dd4hep, but it is entirely possible I missed it.

That's consistent then ;-)

> In any case this just proves how important it is to use the correct units ;)

Absolutely.

tmadlener (approved):

tmadlener:
> - need cross check of efficiency for kink finding

Should this be done before merging, still?

Issue #102, opened 2022-03-11
TOFEstimators seg. faults when fit of the first track chunck fails


dudarboh:
## Issue

As spotted by Uli (@ueinhaus) `TOFEstimators` processor occasionally leads to seg. faults in ~0.001-0.01 of cases.

## Reproducible

Minimal steering file reproducible to recreate the seg fault:
```xml
<marlin>
 <constants>
 </constants>

 <execute>
  <processor name="InitDD4hep" />
  <processor name="MyTOFEstimator_0ps"/>
 </execute>

 <global>
  <parameter name="LCIOInputFiles">
      /pnfs/desy.de/ilc/prod/ilc/mc-opt-3/ild/rec/500-TDR_ws/2f_Z_hadronic/ILD_l5_o1_v02/v02-00-01/00010410/000/rv02-00-01.sv02-00-01.mILD_l5_o1_v02.E500-TDR_ws.I250114.P2f_z_h.eL.pR.n027_149.d_rec_00010410_937.slcio
  </parameter>
  <parameter name="MaxRecordNumber" value="0" />  
  <parameter name="SkipNEvents" value="15" />  
  <parameter name="SupressCheck" value="false" />  
  <parameter name="AllowToModifyEvent" value="true" />
 </global>
    
  <processor name="InitDD4hep" type="InitializeDD4hep">
    <parameter name="DD4hepXMLFile" type="string"> /cvmfs/ilc.desy.de/sw/x86_64_gcc49_sl6/v02-00-01/lcgeo/v00-16-01/ILD/compact/ILD_l5_o1_v02/ILD_l5_o1_v02.xml </parameter>
  </processor>

  <processor name="MyTOFEstimator_0ps" type="TOFEstimators">
  </processor>

</marlin>
```

## Reason for seg. fault

We sort subTrack hits in the direction of the particle movement.
To understand this direction (which hit is first/last) we rely on the last hit of the previous iteration `prevHit` here:

If the fit of the very first  subTrack `(i == 0)` fails, then it is simply skipped with `continue` without pushing any hits into the array which creates a seg. fault on the next iteration as `trackStatesPerHit` is empty and we try to access `prevHit` as `trackStatesPerHit.back()`

https://github.com/iLCSoft/MarlinReco/blob/c7957ec0ab4633eddf545e106f3e03786300c068/TimeOfFlight/src/TOFUtils.cc#L198-L220

## Event example

[...]

PR #103, opened 2022-03-16
Fix seg fault when fit fails and improve subTrack direction check based on z

dudarboh:
BEGINRELEASENOTES

- Try to fit subtrack forward if backward fit fails
- Fix seg. fault if the very first sub track fit fails in both directions
- Change behaviour of checking track direction. Now it is more robust and relies on z position of first and last hits.

ENDRELEASENOTES

This is the fix for #102 with some improved behaviour on how do we check what is the direction of subTrack.

Previous behaviour:
- check position of the last hit from previous iteration `prevHit`
- check **linear** distance between `prevHit` and two edge hits of current iteration.
- choose closest one to be continuation of the prev. subtrack.

Playing with event display has shown it doesn't always work with curly tracks. 
As it may curl in and out of the inner TPC edge so some hits will disappear and the last hit which goes further along the track not necessarily  will be linearly further away from the prev. hit.

Thus i decided to look at the z direction of the track. If track goes along the z, I should know all sub track directions immediately based on their last and first hit z position.
For very perpendicular tracks, if `|zFirstHit - zLastHit| < 5*zTPCResolution` i provide check by the radius, as for the very perpendicular tracks checking by radius should be preferred, although i couldn't find so perfectly perpendicular tracks that z check fails.

### DIFF
--- a/TimeOfFlight/src/TOFEstimators.cc
+++ b/TimeOfFlight/src/TOFEstimators.cc
@@ -98,7 +98,7 @@ void TOFEstimators::init(){
 void TOFEstimators::processEvent(EVENT::LCEvent * evt){
     RandGauss::setTheSeed( marlin::Global::EVENTSEEDER->getSeed(this) );
     ++_nEvent;
-    streamlog_out(MESSAGE)<<"************Event************ "<<_nEvent<<std::endl;
+    streamlog_out(DEBUG9)<<std::endl<<"==========Event========== "<<_nEvent<<std::endl;
     auto startTime = std::chrono::steady_clock::now();
 
     LCCollection* pfos = evt->getCollection(_pfoCollectionName);
@@ -111,6 +111,7 @@ void TOFEstimators::processEvent(EVENT::LCEvent * evt){
 
 
     for (int i=0; i<pfos->getNumberOfElements(); ++i){
+        streamlog_out(DEBUG9)<<std::endl<<"Starting to analyze "<<i+1<<" PFO"<<std::endl;
         ReconstructedParticle* pfo = static_cast <ReconstructedParticle*> ( pfos->getElementAt(i) );
 
         int nClusters = pfo->getClusters().size();
@@ -194,13 +195,14 @@ void TOFEstimators::processEvent(EVENT::LCEvent * evt){
         }
         vector<float> results{float(harmonicMom), float(trackLength), float(timeOfFlight)};
         pidHandler.setParticleID(pfo , 0, 0, 0., algoID, results);
-        streamlog_out(DEBUG6)<<"*****PFO***** "<< i+1<<std::endl;
-        streamlog_out(DEBUG6)<<"momentum: "<< float(harmonicMom)<<" Gev"<<std::endl;
-        streamlog_out(DEBUG6)<<"track length: "<< float(trackLength)<<" mm"<<std::endl;
-        streamlog_out(DEBUG6)<<"time-of-flight: "<< float(timeOfFlight)<<" ns"<<std::endl;
+        streamlog_out(DEBUG9)<<"Final results for the "<<i+1<<" PFO"<<std::endl;
+        streamlog_out(DEBUG9)<<"momentum: "<< float(harmonicMom)<<" Gev"<<std::endl;
+        streamlog_out(DEBUG9)<<"track length: "<< float(trackLength)<<" mm"<<std::endl;
+        streamlog_out(DEBUG9)<<"time-of-flight: "<< float(timeOfFlight)<<" ns"<<std::endl;
+        streamlog_out(DEBUG9)<<std::endl<<std::endl;
     }
 
     auto endTime = std::chrono::steady_clock::now();
     std::chrono::duration<double> duration = endTime - startTime;
-    streamlog_out(DEBUG7)<<"Time spent (sec): "<<duration.count()<<std::endl;
+    streamlog_out(DEBUG9)<<"Time spent (sec): "<<duration.count()<<std::endl;
 }
--- a/TimeOfFlight/src/TOFUtils.cc
+++ b/TimeOfFlight/src/TOFUtils.cc
@@ -198,32 +198,37 @@ std::vector<IMPL::TrackStateImpl> TOFUtils::getTrackStatesPerHit(std::vector<EVE
         int errorFit = MarlinTrk::createFinalisedLCIOTrack(marlinTrk.get(), hits, &refittedTrack, IMarlinTrack::backward, &preFit, bField, maxChi2PerHit);
         //if fit fails, try also fit forward
         if (errorFit != 0){
+            streamlog_out(DEBUG8)<<"Fit backward fails! Trying to fit forward for "<<i+1<<" subTrack in this PFO!"<<std::endl;
+
             marlinTrk.reset( trkSystem->createTrack() );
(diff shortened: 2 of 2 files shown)


tmadlener (commented):
This already looks pretty good to me. One thing that you could do is to replace some of your commends with some debug outputs (`streamlog_out(DEBUG)`) so that it would become possible to get an idea of what is going on from logfiles.

What is the reason for marking this WIP at this point? Only the hardcoded TPC resolution?

dudarboh:
I released WIP flag so, I think it is good to go, if you approve.

I tweaked debug output slightly so it looks more fancy and increased the log to `DEBUG9` so it does not interfere with debug output from sub-processes.

Also looking at the `DDTPCDigiProcessor` code, I realised that actual TPC z resolution is quite small ~ 0.4 - 0.55 mm and another factor is multiple scattering. So it doesn't make sense to make this part of code dependent on TPC z resolution, so I just put plain 10 mm there which I believe is something reasonable..

tmadlener (commented):
Hi @dudarboh,

Thanks for the fixes. I have one comment about the debug levels you have chosen. Is there a reason why you bumped the one that prints the final results for each PFO to `DEBUG9`? I am asking because to me it looks like this should at least be lower than the few `DEBUG8` outputs you have in the function that essentially prints on every subtrack. I would have expected that the summary output should be available at a lower debug level than the more detailed one from the collection of the track states.

dudarboh:
But `DEBUG1` is deeper (more verbose) output than `DEBUG9`...
If I specify `DEBUG1` in the steering file, it will mean print `DEBUG1-9`
If I specify `DEBUG5` in the steering file, it will mean print `DEBUG5-9`
no?

Or I misunderstood your comment?

tmadlener:
> If I specify DEBUG1 in the steering file, it will mean print DEBUG1-9
If I specify DEBUG5 in the steering file, it will mean print DEBUG5-9
no?

It should be the logging level specified in the steering file is the highest that is still printed. So that if you specify, e.g. `DEBUG5` there, it should print `DEBUG1-5` but not `DEBUG6-9` (and it will also print all `MESSAGE`, etc. levels). Is that not the case? I.e. is the behavior you observe different?

dudarboh:
Yes I observe different from what you are saying..

If I specify e.g. `DEBUG9` I will see **only** `DEBUG9`, not `DEBUG1-9`

Output for 1 PFO will be:
```
[ DEBUG9 "MyTOFEstimator_0ps"] ==========Event========== 1
[ DEBUG9 "MyTOFEstimator_0ps"] 
[ DEBUG9 "MyTOFEstimator_0ps"] Starting to analyze 1 PFO
[ DEBUG9 "MyTOFEstimator_0ps"] Final results for the 1 PFO
[ DEBUG9 "MyTOFEstimator_0ps"] momentum: 1.23695 Gev
[ DEBUG9 "MyTOFEstimator_0ps"] track length: 2645.36 mm
[ DEBUG9 "MyTOFEstimator_0ps"] time-of-flight: 8.86054 ns

```

If I specify `DEBUG3`, I will see all the `DEBUG3-9`

Output for 1 PFO will include my `DEBUG8` and `DEBUG9`
with additional ~20 000 lines (for a single PFO) of `DEBUG4` from the fitting:
which has a lot of output like this:

```
[ DEBUG4 "MyTOFEstimator_0ps"]  type 0
[ DEBUG4 "MyTOFEstimator_0ps"]           +-     -+   +-
[ DEBUG4 "MyTOFEstimator_0ps"]           | drho  |   | 0
[ DEBUG4 "MyTOFEstimator_0ps"]           | phi0  |   | 2.705
[ DEBUG4 "MyTOFEstimator_0ps"]  a      = | kappa | = | 0.8888
[ DEBUG4 "MyTOFEstimator_0ps"]           | dz    |   | -1.743e-13
[ DEBUG4 "MyTOFEstimator_0ps"]           | tanl  |   | 0.4555
[ DEBUG4 "MyTOFEstimator_0ps"]           +-     -+   +-
[ DEBUG4 "MyTOFEstimator_0ps"]           +-
[ DEBUG4 "MyTOFEstimator_0ps"]  X0     = | 904
[ DEBUG4 "MyTOFEstimator_0ps"]           | -1528
[ DEBUG4 "MyTOFEstimator_0ps"]           | 950.7
[ DEBUG4 "MyTOFEstimator_0ps"]           +-
[ DEBUG4 "MyTOFEstimator_0ps"]  covMat = +-                                                       -+
[ DEBUG4 "MyTOFEstimator_0ps"]           |       1e+06     -118.2   0.006759     -460.6          0 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |      -118.2      100.2   -0.01135     -38.02          0 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |    0.006759   -0.01135      9.083  -1.14e-06          0 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |      -460.6     -38.02  -1.14e-06      1e+06      119.1 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |           0          0          0      119.1        100 |
[ DEBUG4 "MyTOFEstimator_0ps"]           +-                                                       -+
[ DEBUG4 "MyTOFEstimator_0ps"]  fitered state: 
[ DEBUG4 "MyTOFEstimator_0ps"]  type 1
[ DEBUG4 "MyTOFEstimator_0ps"]           +-     -+   +-
[ DEBUG4 "MyTOFEstimator_0ps"]           | drho  |   | 0
[ DEBUG4 "MyTOFEstimator_0ps"]           | phi0  |   | 2.705
[ DEBUG4 "MyTOFEstimator_0ps"]  a      = | kappa | = | 0.8888
[ DEBUG4 "MyTOFEstimator_0ps"]           | dz    |   | -1.743e-13
[ DEBUG4 "MyTOFEstimator_0ps"]           | tanl  |   | 0.4555
[ DEBUG4 "MyTOFEstimator_0ps"]           +-     -+   +-
[ DEBUG4 "MyTOFEstimator_0ps"]           +-
[ DEBUG4 "MyTOFEstimator_0ps"]  X0     = | 904
[ DEBUG4 "MyTOFEstimator_0ps"]           | -1528
[ DEBUG4 "MyTOFEstimator_0ps"]           | 950.7
[ DEBUG4 "MyTOFEstimator_0ps"]           +-
[ DEBUG4 "MyTOFEstimator_0ps"]  covMat = +-                                                       -+
[ DEBUG4 "MyTOFEstimator_0ps"]           |       1e+06     -118.2   0.006759     -460.6          0 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |      -118.2      100.2   -0.01135     -38.02          0 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |    0.006759   -0.01135      9.083  -1.14e-06          0 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |      -460.6     -38.02  -1.14e-06      1e+06      119.1 |
[ DEBUG4 "MyTOFEstimator_0ps"]           |           0          0          0      119.1        100 |
[ DEBUG4 "MyTOFEstimator_0ps"]           +-                                                       -+
```

If I specify `DEBUG1` i will be way more than 20k lines...

I don't know why it is reversed. And I hope people consistently understood throughout iLCSoft code that `DEBUG1` should be the deepest level...

gaede:
Yes, this is the intended behavior - the VerbosityLevel is a threshold, really - the higher the number the less output...

tmadlener:
OK. Then it seems like I have misunderstood this. I always had thought it works the other way around, where I higher debug level means more verbosity. Anyhow, then I think this looks good.

tmadlener (approved):

PR #104, opened 2022-03-18
Migrating track length code from TOFEstimators into separate processor

dudarboh:
BEGINRELEASENOTES

- Migrated track length and  mean harmonic momentum code into separate processor to save CPU computing time

ENDRELEASENOTES

TODOs:

- [x] Finish migrating and cleaning up all relevant code to the TrackLengthProcessor
- [x] Do clean up of the TOFEstimators leaving only TOF related methods and output
- [x] Test the results with some example script, which uses both processors `TOFEstimators` and `TrackLengthProcessor` to calculate the mass distributions.
- [x] Update the documentation of both processors accordingly

PR #105, opened 2022-03-28
[WIP] Update Doxygen documentation for MarlinReco packages

dudarboh:
BEGINRELEASENOTES

- Add paths of all MarlinReco processors to the INPUT of doxygen
- Enable source browser and modify few style settings to match ROOT format
- Document existing processors into separate categories
- Minor clean up of documentation

ENDRELEASENOTES

TODO:

- [ ] Create subgroups of documentation for separate processors
- [ ] Add existing README subpages into documentation
- [ ] Remove documentation of empty namespaces, etc


dudarboh:
Hi @tmadlener,

At this point I would consider this a basic minimal PR. It just adds paths of processors for documentation.

But as you can see from TODO list, there still few things I would like to add, to make it look fancier than just a list of classes.
But these would require a bit more changes in many files, so just in case if you won't be happy with those changes we can always come back to this commit in the end.

tmadlener:
Thanks @dudarboh. I have no particularly strong feelings about the style or the overall organization. I just saw that you have tweaked quite a few doxygen defaults. Do you happen to have the current results of running doxygen on MarlinReco available somewhere? I think it might be easier to look at the result, rather than sifting through doxygen configuration and guessing what it might look like in the end. There was a link in #104 somewhere, is that pointing to the current version? (Doesn't really matter if that PR is part of the output).

dudarboh:
@tmadlener
For debugging purposes, I will host documentation from current PR here:
https://www.desy.de/~dudarboh/marlinreco_doc/html/index.html

PR #106, opened 2022-10-21
hot fix seg. fault when there is not enough hits to calculate the track length

dudarboh:
BEGINRELEASENOTES

- Fix a seg. fault, in rare cases, when the track fit fails in both directions due to the lack of hits.

ENDRELEASENOTES

Found out that around 2.2% of my jobs seg. faulted because of that. Few cases, which i encountered, had tracks with 3-6 hits in total.
I am not sure, how the track successfully created in the first place, as I have tried to keep fit procedure as similar to the production code as possible.


tmadlener:
gcc CI builds seem to pick up the wrong version of LCIO during linking. iLCSoft/LCIO#150 has removed `virtual` from all `LCRelationNavigator` functionality.

tmadlener:
Same as for #107, will wait for the key4hep-stack to pick up the newest version of LCIO and make sure things build there before merging. I will rebase this once #107 has been merged.

tmadlener:
merging, since key4hep-nightlies based workflow is happy.

PR #107, opened 2022-11-09
Remove no longer available CI workflow based on gcc8

tmadlener:
BEGINRELEASENOTES
- Remove no longer available CI workflow based on gcc8 since the underlying nightly builds are no longer available
- Update github actions to latest available versions

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -9,9 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'iLCSoft/MarlinReco'
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


tmadlener:
I am fairly certain the failure on gcc10 is due to an older version of LCIO being picked up by the linker at some point in the process. I debugged this a bit, but could not really find where the linker picks this up. I will wait for a new nightly build of the key4hep stack, after key4hep/key4hep-stack#441 has been merged, to make sure things build also with gcc before merging this.

tmadlener:
Merging this, as the key4hep-nightlies based build works with the latest version of LCIO.

PR #112, opened 2023-05-24
Update of the Track length calculator

dudarboh:
BEGINRELEASENOTES

- Switch to the helix formula without Omega: $\ell_{i} = \frac{|z_{i+1} - z_{i}|}{|\tan{\lambda_{i}}|}\sqrt{1 +\tan^2{\lambda_{i}}}$. It shows the best performance so far.
- Other bug fixes and consistency improvements.

ENDRELEASENOTES

+ The most major change is the new formula to calculate the helical distance between two neighboring track states.
We change from $\ell_{i} = \sqrt{\left( \frac{\varphi_{i+1} - \varphi_{i}}{\Omega_{i}}\right)^{2} + \left( z_{i+1} - z_{i} \right)^{2} }$ to $\ell_{i} = \frac{|z_{i+1} - z_{i}|}{|\tan{\lambda_{i}}|}\sqrt{1 +\tan^2{\lambda_{i}}}$ as it shows much more precise results for all tracks.
+ Fixed a bug where the track state at the calorimeter surface was not included when track fit of the last subTrack has failed. Now we extrapolate to the calorimeter from the latest successfuly fitted subTrack if PFO has any matched shower in the calorimeter.
+ Fixed a bug where an additional extra track state was added by mistake before the calorimeter track state.
+ Improved robustness of identifying first TPC curl subTrack. Now it is expected to work100% of the time, not most of the time.
+ Now track length calculated for all PFOs which have at least one track, independently of number of clusters, as it should be. In case of the PFOs with 2+ tracks attached (e.g. kinks, V0s) track length takes only the first track to do the calculation. For these cases track length might not be very well defined.
+ Code cleanup, style and comments improvements and clarifications.

### DIFF
--- a/Tracking/TrackLength/src/TrackLengthProcessor.cc
+++ b/Tracking/TrackLength/src/TrackLengthProcessor.cc
@@ -68,12 +68,9 @@ void TrackLengthProcessor::processEvent(EVENT::LCEvent * evt){
         streamlog_out(DEBUG9)<<std::endl<<"Starting to analyze "<<i+1<<" PFO"<<std::endl;
         ReconstructedParticle* pfo = static_cast <ReconstructedParticle*> ( pfos->getElementAt(i) );
 
-        int nClusters = pfo->getClusters().size();
-        int nTracks = pfo->getTracks().size();
-
-        if( nTracks == 0 ){
-            // Calculate track length only for particles with a track.
-            //Note: If pfo has more than one track attached, only the first track will be used. The results may be not accurate in such cases. 
+        if( pfo->getTracks().empty() ){
+            // Calculate track length only for particles with at least one track.
+            // Note: If the pfo has more than one track attached (e.g. kink, v0), only the first track will be used. The results may be not accurate in these cases. 
             vector<float> results{0., 0., 0., 0.};
             pidHandler.setParticleID(pfo , 0, 0, 0., algoID, results);
             continue;
--- a/Tracking/TrackLength/src/TrackLengthProcessor.cc
+++ b/Tracking/TrackLength/src/TrackLengthProcessor.cc
@@ -78,8 +78,7 @@ void TrackLengthProcessor::processEvent(EVENT::LCEvent * evt){
         }
         Track* track = pfo->getTracks()[0];
 
-        vector<Track*> subTracks = getSubTracks(track);
-        vector<TrackStateImpl> trackStates = getTrackStatesPerHit(subTracks, _trkSystem, _bField);
+        vector<TrackStateImpl> trackStates = getTrackStates(pfo, _trkSystem, _bField);
 
         double trackLengthToSET = 0.;
         double harmonicMomToSET = 0.;
--- a/Tracking/TrackLength/src/TrackLengthProcessor.cc
+++ b/Tracking/TrackLength/src/TrackLengthProcessor.cc
@@ -71,6 +71,7 @@ void TrackLengthProcessor::processEvent(EVENT::LCEvent * evt){
         if( pfo->getTracks().empty() ){
             // Calculate track length only for particles with at least one track.
             // Note: If the pfo has more than one track attached (e.g. kink, v0), only the first track will be used. The results may be not accurate in these cases. 
+            streamlog_out(DEBUG9)<<"PFO has no tracks. Writing zeros."<<std::endl;
             vector<float> results{0., 0., 0., 0.};
             pidHandler.setParticleID(pfo , 0, 0, 0., algoID, results);
             continue;
--- a/Tracking/TrackLength/src/TrackLengthProcessor.cc
+++ b/Tracking/TrackLength/src/TrackLengthProcessor.cc
@@ -10,6 +10,7 @@
 #include "marlinutil/GeometryUtil.h"
 #include "MarlinTrk/Factory.h"
 #include "EVENT/SimTrackerHit.h"
+#include "UTIL/TrackTools.h"
 
 using namespace TrackLengthUtils;
 using std::vector;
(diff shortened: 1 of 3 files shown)


tmadlener:
I just had a brief look (so no detailed comments yet). There is a lot of removed code / functionality. I suppose this was mainly useful for development and testing and now that things have settled we really don't want to offer that functionality any longer since the helix length formula will be the only one that is used in the future? Otherwise, we could consider keeping the "legacy" functionality around in case one wants to redo these studies in the future again.

dudarboh:
Indeed, I think I could keep few utility functions, without using them in the code itself, so that if we want to test all different options again, one at least has all the utils here, rather than deducing them mathematically again from scratch... I will add the onse I removed and one extra next week

dudarboh:
As discussed on the meeting today, we postpone the idea of calculating track length from the vertex position, as: a) for primary vertices, this should not make any difference, as (0,0,0) is just a reference point and distance to the PrimVertex accounted in z0 paramter. b) for secondary vertices, we keep it as it is, as for the current time-of-flight studies we always use time of the event, as the initial time, not the time of the secondary vertex production. So this needs a more involved treatment, even if needed.

dudarboh:
As for the track state at the last hit discussion and the difference between `getTrackState()` vs `propagate()`.
I would prefer to use `getTrackState()` for now for all the hits. As it seems to show reasonable results as long as the track length calculation is concerned and code looks cleaner.
We can always implement this (probably) sub-percent level of improvement in the future.

Tracking/TrackLength/src/TrackLengthProcessor.cc
```diff
@@ -66,19 +68,14 @@ void TrackLengthProcessor::processEvent(EVENT::LCEvent * evt){
...
         ReconstructedParticle* pfo = static_cast <ReconstructedParticle*> ( pfos->getElementAt(i) );
 
-        int nClusters = pfo->getClusters().size();
-        int nTracks = pfo->getTracks().size();
-
-        if( nClusters != 1 || nTracks != 1){
-            // Analyze only simple pfos. Otherwise write dummy zeros
+        if( pfo->getTracks().empty() ){
+            // Calculate track length only for particles with at least one track.
+            // Note: If the pfo has more than one track attached (e.g. kink, v0), only the first track will be used. The results may be not accurate in these cases.
```
tmadlener, line 73:
```suggestion
            // Note: If the pfo has more than one track attached (e.g. kink, v0), only the first track will be used. The results may be not accurate in these cases. 
            streamlog_out(DEBUG9) << "PFO has no tracks. Writing zeros." << std::endl;
```

Might be a useful debug message to have.

Tracking/TrackLength/src/TrackLengthUtils.cc
```diff
@@ -46,86 +45,85 @@ IMPL::TrackStateImpl TrackLengthUtils::getTrackStateAtHit(MarlinTrk::IMarlinTrac
...
-    //Read documentation in the header file for details.
-    int startIdx;
-    if( std::abs(nTPCHits - nSubTrack0Hits) <= 1  ) startIdx = 1;
-    else if ( std::abs(nTPCHits - nSubTrack1Hits) <= 1 ) startIdx = 2;
-    else{
-        //FIXME: This happens very rarily (0.01%) for unknown reasons, so we just, skip adding subTracks...
-        streamlog_out(WARNING)<<"Can't understand which subTrack is responsible for the first TPC hits! Skip adding subTracks."<<std::endl;
-        return subTracks;
+    auto isTPCHit = [](TrackerHit* hit) -> bool {
+        UTIL::BitField64 encoder( UTIL::LCTrackerCellID::encoding_string() ) ;
```
tmadlener, line 94: Similar comment as already for the TOFUtils. The construction of the `encoder` should probably be outside the lambda, as I am not entirely sure whether the compiler is clever enough to figure out this optimization opportunity.

Tracking/TrackLength/src/TrackLengthUtils.cc
```diff
@@ -46,86 +45,85 @@ IMPL::TrackStateImpl TrackLengthUtils::getTrackStateAtHit(MarlinTrk::IMarlinTrac
...
+    };
+
+    int indexOfFirstTPCCurl = 0;
+    for(int i = 0; i < nSubTracks; ++i){
+        Track* subTrack = track->getTracks()[i];
+        auto hits = subTrack->getTrackerHits();
+        if ( std::find_if(hits.begin(), hits.end(), isTPCHit) != hits.end() ){
+            indexOfFirstTPCCurl = i;
+            break;
+        }
```
tmadlener, line 107: This seems to be (exactly?) the same piece of code as in TOFUtils and as far as I can tell the main purpose of these lines is to figure out which track has the first complete curl in the TPC. Maybe we could consider moving this to the TrackUtils (in a future PR, or in this).

Tracking/TrackLength/src/TrackLengthProcessor.cc
```diff
@@ -66,19 +68,16 @@ void TrackLengthProcessor::processEvent(EVENT::LCEvent * evt){
...
-        int nClusters = pfo->getClusters().size();
-        int nTracks = pfo->getTracks().size();
-
-        if( nClusters != 1 || nTracks != 1){
-            // Analyze only simple pfos. Otherwise write dummy zeros
+        if( pfo->getTracks().empty() ){
+            // Calculate track length only for particles with at least one track.
+            // Note: If the pfo has more than one track attached (e.g. kink, v0), only the first track will be used. The results may be not accurate in these cases. 
+            streamlog_out(DEBUG9) << "PFO has no tracks. Writing zeros." << std::endl;
+            streamlog_out(DEBUG9)<<"PFO has no tracks. Writing zeros."<<std::endl;
```
tmadlener, line 75:
```suggestion
```
Ah sorry. Just saw that this was already there. Removing again.

PR #113, opened 2023-05-25
Update of the time of flight processor

dudarboh:
BEGINRELEASENOTES

- Always use the first (before)  --> latest (now) curl in the track to get extrapolated track position at the calorimeter surface. This gives sometimes better estimate of the track position at the ECAL surface, especially for the tracks with large number of curls. Thus  new version gives better time of flight correction for the distance to the surface and thus TOF itself.
- Minor style improvements

ENDRELEASENOTES

### DIFF
--- a/TimeOfFlight/include/TOFUtils.h
+++ b/TimeOfFlight/include/TOFUtils.h
@@ -13,15 +13,9 @@
 */
 namespace TOFUtils{
 
-    /** Returns TPC outer radius from the DD4hep detector geometry.
+    /** Returns SET hit if exists.
     */
-    double getTPCOuterR();
-
-    /** Returns SET hit.
-    Checks the \f$ \rho = \sqrt{x^{2}+y^{2}} \f$ of the tracker hit.
-    If \f$ \rho > R_{\mathrm{TPC, outer}} \f$  then it is the SET hit.
-    */
-    EVENT::TrackerHit* getSETHit(EVENT::Track* track, double tpcOuterR);
+    EVENT::TrackerHit* getSETHit(EVENT::Track* track);
 
     /** Get a subset of the cluster calorimeter hits.
     Returns the subset of the cluster calorimeter hits which we call "Frank"
--- a/TimeOfFlight/include/TOFUtils.h
+++ b/TimeOfFlight/include/TOFUtils.h
@@ -13,11 +13,6 @@
 */
 namespace TOFUtils{
 
-    /** Get track momentum at the track state.
-    Returns momentum Vector3D from the given track state.
-    */
-    dd4hep::rec::Vector3D getHelixMomAtTrackState(const EVENT::TrackState& ts, double bField);
-
     /** Returns TPC outer radius from the DD4hep detector geometry.
     */
     double getTPCOuterR();
--- a/TimeOfFlight/include/TOFUtils.h
+++ b/TimeOfFlight/include/TOFUtils.h
@@ -61,7 +61,7 @@ namespace TOFUtils{
     /** Get TrackState of the track at the Calorimeter surface. In case of multiple curls take the latest curl (closest to the calorimeter)
     to get the extrapolated position.
     */
-    const EVENT::TrackState* geTrackStateAtCalorimeter(EVENT::Track* track);
+    const EVENT::TrackState* getTrackStateAtCalorimeter(EVENT::Track* track);
 
 
 }
--- a/TimeOfFlight/include/TOFUtils.h
+++ b/TimeOfFlight/include/TOFUtils.h
@@ -64,6 +64,12 @@ namespace TOFUtils{
     */
     double getTofFrankFit( std::vector<EVENT::CalorimeterHit*> selectedHits, EVENT::Track* track, double timeResolution);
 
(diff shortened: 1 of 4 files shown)


TimeOfFlight/src/TOFUtils.cc
```diff
@@ -28,42 +29,54 @@ using dd4hep::rec::Vector3D;
...
-
-double TOFUtils::getTPCOuterR(){
-    auto& detector = Detector::getInstance();
-    DetElement tpcDet = detector.detector("TPC");
-    FixedPadSizeTPCData* tpc = tpcDet.extension <FixedPadSizeTPCData>();
-    return tpc->rMaxReadout/dd4hep::mm;
+EVENT::TrackerHit* TOFUtils::getSETHit(EVENT::Track* track){
+    vector<TrackerHit*> hits = track->getTrackerHits();
+    auto isSETHit = [](TrackerHit* hit) -> bool {
+        UTIL::BitField64 encoder( UTIL::LCTrackerCellID::encoding_string() ) ;
```
tmadlener, line 35: I am not entirely sure (would need to be profiled / benchmarked), but I am not sure if the compiler sees this as something that he can inline (and in the end hoist outside the loop), so that he doesn't construct an encoder in each call in `find_if` below. Without benchmarking, the easiest thing to do probably would be to create the encoder outside the lambda and simply capture it by reference.

TimeOfFlight/src/TOFUtils.cc
```diff
@@ -28,42 +29,54 @@ using dd4hep::rec::Vector3D;
...
+        return subdet == UTIL::ILDDetID::SET;
+    };
+    auto it = std::find_if(hits.begin(), hits.end(), isSETHit);
+    if ( it != hits.end() ) return *it;
+    return nullptr;
 }
 
+const EVENT::TrackState* TOFUtils::geTrackStateAtCalorimeter(EVENT::Track* track){
+    auto isTPCHit = [](TrackerHit* hit) -> bool {
+        UTIL::BitField64 encoder( UTIL::LCTrackerCellID::encoding_string() ) ;
```
tmadlener, line 47: Same here.

TimeOfFlight/src/TOFUtils.cc
```diff
@@ -28,42 +29,54 @@ using dd4hep::rec::Vector3D;
...
-    double phi = ts.getPhi();
-    double d0 = ts.getD0();
-    double z0 = ts.getZ0();
-    double omega = ts.getOmega();
-    double tanL = ts.getTanLambda();
-
-    HelixClass helix;
-    helix.Initialize_Canonical(phi, d0, z0, omega, tanL, bField);
-    return helix.getMomentum();
-}
```
tmadlener, line 42: Does removing this break someone elses code?

tmadlener (commented):
On the physics I can't judge, so I will believe you ;) A few comments below. 

Could you also put the sentence that is currently outside the release notes as a sub bullet to the first point in the release notes? I think that is quite useful information.

TimeOfFlight/src/TOFUtils.cc
dudarboh: I don't think so. I think we are safe to remove this.
In the worst case, if somebody have used this specific function from TOFUtils in their local code for some reason, instead of two existing UTIL functions (in LCIO and MarlinUtil), which do the same thing, I think it is a good insentive to switch to UTIL functions. Also, this code is present in dozen of places throughout many packages of MarlinReco, so I would be happy to remove another copy of it, just nicely put in the util function...

TimeOfFlight/include/TOFUtils.h
```diff
@@ -69,6 +58,12 @@ namespace TOFUtils{
     */
     double getTofFrankFit( std::vector<EVENT::CalorimeterHit*> selectedHits, EVENT::Track* track, double timeResolution);
 
+    /** Get TrackState of the track at the Calorimeter surface. In case of multiple curls take the latest curl (closest to the calorimeter)
+    to get the extrapolated position.
+    */
+    const EVENT::TrackState* geTrackStateAtCalorimeter(EVENT::Track* track);
```
tmadlener, line 64:
```suggestion
    const EVENT::TrackState* getTrackStateAtCalorimeter(EVENT::Track* track);
```
typo I missed previously.

TimeOfFlight/src/TOFUtils.cc
```diff
@@ -28,42 +29,54 @@ using dd4hep::rec::Vector3D;
...
+    }
 
-EVENT::TrackerHit* TOFUtils::getSETHit(EVENT::Track* track, double tpcOuterR){
-    vector<TrackerHit*> hits = track->getTrackerHits();
-    TrackerHit* lastHit = hits.back();
-    Vector3D pos (lastHit->getPosition());
-
-    if ( pos.rho() > tpcOuterR ) return lastHit;
-    return nullptr;
+    // Take the trackState at thhe calorimeter surface always from the latest curl
```
tmadlener, line 64:
```suggestion
    // Take the trackState at the calorimeter surface always from the latest curl
```

tmadlener (approved):

PR #114, opened 2023-06-07
Add LeptonIDProcessor

Zehvogel:
Here the (long awaited) draft for the LeptonIDProcessor I used for my master thesis results. 
There is still some cleaning up to do that I will try to get in this week.

At the moment there is still a PIDUtil.{h,cc} which will also used by my SLDIDProcessor and might be better off in MarlinUtil. I can move it there if necessary. The SLDIDProcessor still needs some more cleaning up though before I will open another PR for that one.

BEGINRELEASENOTES
- Added `LeptonIDProcessor` to identify electrons and muons in jets using boosted decision trees.

ENDRELEASENOTES

### DIFF
--- a/Analysis/LeptonID/src/LeptonIDProcessor.cc
+++ b/Analysis/LeptonID/src/LeptonIDProcessor.cc
@@ -2,6 +2,7 @@
 #include "PIDutil.h"
 #include <iostream>
 #include <cmath>
+#include <optional>
 
 #include <EVENT/LCCollection.h>
 #include <EVENT/ReconstructedParticle.h>
@@ -219,13 +220,11 @@ void LeptonIDProcessor::modifyEvent(LCEvent *evt)
         return;
     }
 
-    //LCRelationNavigator RecoMCTruthNavigator(RecoMCTruthLink);
-    //LCRelationNavigator MCTruthRecoNavigator(MCTruthRecoLink);
-    std::unique_ptr<LCRelationNavigator> RecoMCTruthNavigator;
-    std::unique_ptr<LCRelationNavigator> MCTruthRecoNavigator;
+    std::optional<LCRelationNavigator> RecoMCTruthNavigator;
+    std::optional<LCRelationNavigator> MCTruthRecoNavigator;
     if (_buildTree) {
-        RecoMCTruthNavigator = std::make_unique<LCRelationNavigator>(RecoMCTruthLink);
-        MCTruthRecoNavigator = std::make_unique<LCRelationNavigator>(MCTruthRecoLink);
+        RecoMCTruthNavigator = LCRelationNavigator(RecoMCTruthLink);
+        MCTruthRecoNavigator = LCRelationNavigator(MCTruthRecoLink);
     }
 
     PIDHandler pidh(PandoraPFOs);


Analysis/LeptonID/include/PIDutil.h
```diff
@@ -0,0 +1,33 @@
...
+#include <EVENT/Vertex.h>
+#include <UTIL/LCRelationNavigator.h>
+#include <marlinutil/WeightedPoints3D.h>
+#include <TVector3.h>
+
+using namespace lcio;
+
+float getCorrectdEdx(float dEdx, Track *track);
+
+LCObject *getRelated(LCObject *p, LCRelationNavigator const &nav, int &_cw, int &_tw);
```
tmadlener, line 14:
```suggestion
std::tuple<LCObject*, float, float> getRelated(LCObject *p, LCRelationNavigator const &nav);
```
Would be a possibility to get rid of the output reference parameters. Since we are using c++17 in any case it is also easy to destructure this with structured bindings on the calling site:
```cpp
/*const*/ auto [obj, track_w, cluster_w] = getRelated(/* */);
```


In general, it looks like the weights should be `float` and not `int` unless you specifically want them to be in permille. (You seem to indicate that in the implementation).

Analysis/LeptonID/src/LeptonIDProcessor.cc
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    //LCRelationNavigator RecoMCTruthNavigator(RecoMCTruthLink);
+    //LCRelationNavigator MCTruthRecoNavigator(MCTruthRecoLink);
+    std::unique_ptr<LCRelationNavigator> RecoMCTruthNavigator;
+    std::unique_ptr<LCRelationNavigator> MCTruthRecoNavigator;
+    if (_buildTree) {
+        RecoMCTruthNavigator = std::make_unique<LCRelationNavigator>(RecoMCTruthLink);
+        MCTruthRecoNavigator = std::make_unique<LCRelationNavigator>(MCTruthRecoLink);
+    }
```
tmadlener, line 229: Not sure I immediately see why these cannot live on the stack. What am I missing?

Analysis/LeptonID/src/LeptonIDProcessor.cc
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        auto clusters = p->getClusters();
+        int nClusters = clusters.size();
+        if (nClusters != 1) {
+            continue;
+        }
+
+        if (_buildTree) {
+            // TODO: maybe use PIDUtil getMCParticle instead
```
tmadlener, line 261: Why not do it?

Analysis/LeptonID/src/LeptonIDProcessor.cc
```diff
@@ -0,0 +1,374 @@
...
+        }
+        clear();
+    }
+}
+
+void LeptonIDProcessor::check(LCEvent *evt)
+{
+    // nothing to check here - could be used to fill checkplots in reconstruction processor
+    (void) evt;
+}
```
tmadlener, line 343:
```suggestion
void LeptonIDProcessor::check(LCEvent *)
{
    // nothing to check here - could be used to fill checkplots in reconstruction processor
}
```
Easier way to get rid of unused variable warnings. ;)

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,374 @@
...
+
+#include <marlin/VerbosityLevels.h>
+
+#include <EVENT/ReconstructedParticle.h>
+
+float getCorrectdEdx(float dEdx, Track *track)
+{
+    // dEdx correction stolen from the dEdx correction processor
+    double lambda = abs(atan(track->getTanLambda()) * 180.0 / M_PI);
+    std::vector<double> par = {0.970205, 0.0007506, 4.41781e-8, 5.8222e-8};
```
tmadlener, line 14: Is there any way to re-use that existing functionality? Otherwise this is a prime candidate to get out of sync given the hardcoded values here.

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,374 @@
...
+
+    double f3 = 1 / (par[0] + par[1] * lambda + par[2] * pow(lambda, 2) + par[3] * pow(lambda, 3));
+
+    float dEdx_v2 = dEdx * f3;
+    return dEdx_v2;
+}
+
+LCObject *getRelated(LCObject *p, LCRelationNavigator const &nav, int &_cw, int &_tw)
```
tmadlener, line 22: If I understand this correctly, this function could almost entirely be replaced by new functionality in the [`LCRelationNavigator`](https://github.com/iLCSoft/LCIO/blob/ed1a20027baf614d53758c8dd279411749938267/src/cpp/include/UTIL/LCRelationNavigator.h#L85-L120) and [`MarlinUtil`](https://github.com/iLCSoft/MarlinUtil/blob/402540c525b98688e098ddbf2f3957eb9148d2f0/source/include/MarlinUtil.h#L64-L80)

Analysis/LeptonID/include/PIDutil.h
```diff
@@ -0,0 +1,33 @@
...
+
+WeightedPoints3D getWeightedPoints3D(const Cluster * clu, const LCCollection *PandoraClusters = nullptr);
+
+MCParticle *getMCParent(const MCParticle *mcp);
+
+std::vector<float> getPosMomFromTrackState(const TrackState *trackstate, double B);
```
tmadlener, line 20: This seems to be unused in this PR(?)

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,374 @@
...
+        zhits.reserve(n_hits);
+        for (const auto *hit: hits) {
+            ehits.push_back(hit->getEnergy());
+            auto pos = hit->getPosition();
+            xhits.push_back(pos[0]);
+            yhits.push_back(pos[1]);
+            zhits.push_back(pos[2]);
+        }
+        auto res = WeightedPoints3D(int(n_hits), ehits.data(), xhits.data(), yhits.data(), zhits.data());
+        return res;
```
tmadlener, line 84:
```suggestion
        return WeightedPoints3D(int(n_hits), ehits.data(), xhits.data(), yhits.data(), zhits.data());
```
No need to rob the compiler of an an easy opportunity to do RVO. ;)

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,374 @@
...
+}
+
+double intersectTrackLine(Track *track, TVector3 momentumOfLine,
+                          std::vector<double> pointOnLine,
+                          Vertex *primaryVertex, TVector3 &PCAatTrack,
+                          TVector3 &PCAatLine)
+{
+    double phi_min, phi_max, v_min, v_max;
+    phi_min = track->getPhi() - 3.14159265359 / 40.0;
+    phi_max = track->getPhi() + 3.14159265359 / 40.0;
```
tmadlener, line 219: Can we give the magic number `40` a name here?

Analysis/LeptonID/include/PIDutil.h
```diff
@@ -0,0 +1,33 @@
...
+
+double intersectTrackLine(Track *track, TVector3 momentumOfLine,
```
tmadlener, line 22: Similar comment here about returning a `tuple` instead of having out parameters.

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,374 @@
...
+        double dist_z_sq = (vtxPos[2] - primVtxPos[2]) * (vtxPos[2] - primVtxPos[2]);
+
+        double dist = sqrt(dist_x_sq + dist_y_sq + dist_z_sq);
+
+        if (dist < minDist)
+        {
+            minDist = dist;
+            closestVtx = vtx;
+        }
+    }
```
tmadlener, line 347: This is another candidate for `std::min_element` and a lambda which would make the intent immediately clear. That would also eliminate the need for an explicit `minDist` variable. Additionally the `sqrt` is unnecessary here, since we are only interested in how the distances compare not their actual values.

tmadlener (commented):
Thanks for preparing this. I don't have a strong opinion on whether the `PIDUtils` should live in `MarlinUtil`. If there is a need for that we can always move them later. This looks good in general, I effectively have only two general comments and a few smaller ones, mainly about using (rather new) outside functionality to achieve the same things.

The general ones:
- The weights file could become part of ILDConfig, where such things usually live. We would need to find a place there. We might postpone that until this becomes part of a standard reconstruction workflow, unless @gaede wants to move this already now.
- Could you add (brief) docstrings to the functionality in `PIDUtils.h`? For some of them it is quite hard to guess from the signature alone what they are going to return.

Regarding making the return types of the functions with out parameters `tuple`s instead. That is more of a personal preference, and I have the feeling not everybody knows about this opportunity yet, so I usually take it as an opportunity to point it out. You can keep them as they are if you want (and add docstrings).

Analysis/LeptonID/xml/test_eval.xml
```diff
@@ -0,0 +1,93 @@
...
+  <!--verbosity level of this processor ("DEBUG0-4,MESSAGE0-4,WARNING0-4,ERROR0-4,SILENT")-->
+  <!--parameter name="Verbosity" type="string">DEBUG </parameter-->
+</processor>
+
+<processor name="MyLeptonIDProcessor" type="LeptonIDProcessor">
+   <!--verbosity level of this processor ("DEBUG0-4,MESSAGE0-4,WARNING0-4,ERROR0-4,SILENT")-->
+   <parameter name="Verbosity" type="string">DEBUG </parameter>
+   <parameter name="BuildTree" type="bool">false</parameter>
+   <parameter name="EvalMVA" type="bool">true</parameter>
+   <parameter name="weightfile">/afs/desy.de/group/flc/pool/reichenl/cleanup/MarlinReco/Analysis/LeptonID/training/PID_multi_jet_dEdx_800t_3d_cm50r50/weights/PID_multi_jet_dEdx_800t_3d_cm50r50_BDTG.weights.xml</parameter>
```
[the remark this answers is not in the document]
Zehvogel: yes!

Analysis/LeptonID/src/LeptonIDProcessor.cc
Zehvogel: didn't know that one yet, thanks!

Analysis/LeptonID/src/LeptonIDProcessor.cc
Zehvogel: Just need to refactor `getMCParticle` a bit for this to be able to pass the weight cuts, but now is the right moment to do that so I will :)

Analysis/LeptonID/src/PIDutil.cc
Zehvogel: Good idea, these weren't there when I wrote this.

This would be straightforward if [`getMaxWeightIdx`](https://github.com/iLCSoft/LCIO/blob/ed1a20027baf614d53758c8dd279411749938267/src/cpp/include/UTIL/LCRelationNavigator.h#L27-L35) was `public`. Even then, implementing the same logic would call it twice to loop over the weights. If I use `getRelatedToMaxWeightObject` and `getRelatedToMaxWeight` to emulate the same logic we will loop over the weights four times. Not sure if this will have any meaningful impact but for me it feels wrong to do that.

tl; dr: I would prefer to have a `getRelatedToMaxWeightIndex` function in `LCRelationNavigator` to do this.. Can we still get that in even though you just did a release @tmadlener?

Analysis/LeptonID/src/PIDutil.cc
Zehvogel: Maybe. This whole function is more or less copy paste [from Yasser](https://github.com/yradkhorrami/SLDecayCorrection/blob/58f59925bc893cfe1abb17fa9a6bbac48b00af76/src/flightDirection.cc#L282-L436). But I think it is also not directly used in this PR.

Zehvogel:
Thanks for the detailed feedback and the helpful suggestions Thomas! 

I will cleanup `PIDUtil` to only include things actually needed in this PR. A lot of the functions in there will only be needed by the SLDIDProcessor and I don't know yet when I will get around to upload that one.

Regarding the weights I don't care where they live. I don't think that you want to use the same ones for a mass production, they could use some refinement. I only included them to make it easier to try out this processor. I also noticed now that they are horribly big... not sure if we should keep them in here :/

Analysis/LeptonID/src/PIDutil.cc
Zehvogel: This can just be removed, it was in here for yet another processor that just made some n-tuples...

Analysis/LeptonID/src/LeptonIDProcessor.cc
Zehvogel: Here my intention was to only initialize the `LCRelationNavigator` only if necessary to not depend on the existence of `MCParticle` and relation information when not needed. I first tried something like

```c++
LCRelationNavigator RecoMCTruthNavigator;
LCRelationNavigator MCTruthRecoNavigator;
  if (_buildTree) {
    RecoMCTruthNavigator = RecoMCTruthNavigator(RecoMCTruthLink);
    MCTruthRecoNavigator = MCTruthRecoNavigator(MCTruthRecoLink);
  }
```
but that doesn't work. This did but I am open to suggestions.. :)

tmadlener: Ah right, protected default constructor will do that.

Analysis/LeptonID/src/PIDutil.cc
tmadlener: `getMaxWeightIdx` is a free function in the `UTIL` namespace, it should be possible to call that just passing in the corresponding weights, [similar to what is done in the LCRelationNavigator](https://github.com/iLCSoft/LCIO/blob/ed1a20027baf614d53758c8dd279411749938267/src/cpp/include/UTIL/LCRelationNavigator.h#L117-L119).

I do agree that having something that returns the object and the weight would be useful in your case. No problem adding that (even if we just had a new tag).

Zehvogel: Thx, must have overlooked that. I am using `getMaxWeightIdx` now. I can add a method that returns object and weight to the LCRelationNavigator later and update PIDUtil in a future PR. I am happy enough with this for now :)

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,137 @@
...
+#include <set>
+
+#include <marlin/VerbosityLevels.h>
+
+#include <marlinutil/MarlinUtil.h>
+
+#include <EVENT/ReconstructedParticle.h>
+
+std::tuple<LCObject*, float, float> getRelated(LCObject* p, LCRelationNavigator const& nav) {
+  auto  weights = nav.getRelatedToWeights(p);
```
tmadlener, line 12:
```suggestion
  const auto&  weights = nav.getRelatedToWeights(p);
```

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,137 @@
...
+    // too messy, skip
+    return std::make_tuple(nullptr, 0.0F, 0.0F);
+  }
+
+  auto objects = nav.getRelatedToObjects(p);
+  return std::make_tuple(objects[max_ci], max_cw, max_tw);
+}
+
+WeightedPoints3D getWeightedPoints3D(const Cluster* clu, const LCCollection* PandoraClusters) {
+  auto   hits   = clu->getCalorimeterHits();
```
tmadlener, line 32:
```suggestion
  const auto&   hits   = clu->getCalorimeterHits();
```

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    std::vector<double> cov_fr_clu;
+    for (int i = 0; i < 3; i++) {
+      for (int j = 0; j < 3; j++) {
+        cov_fr_clu.push_back(seen_covmat[i][j]);
+      }
+    }
+
+    auto                dir_error = clu->getDirectionError();
```
tmadlener, line 95:
```suggestion
    const auto&                dir_error = clu->getDirectionError();
```

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,137 @@
...
+    std::vector<double> thphcov(dir_error.begin(), dir_error.end());
+
+    return WeightedPoints3D(cogv, cov_fr_clu, thphcov, npoints, wgt_sum, wgt2_sum, wgt4_sum);
+  }
+}
+
+MCParticle* getMCParent(const MCParticle* mcp) {
+  MCParticle* parent = nullptr;
+
+  auto parents = mcp->getParents();
```
tmadlener, line 105:
```suggestion
  const auto& parents = mcp->getParents();
```

Analysis/LeptonID/src/LeptonIDProcessor.cc
```diff
@@ -0,0 +1,356 @@
...
+    {
+        streamlog_out(DEBUG) << "pfo loop: " << i << std::endl;
+        auto *p = dynamic_cast<ReconstructedParticle *>(InputPFOs->getElementAt(i));
+
+        // We only want to look at charged particles with exactly one track and cluster for now
+        if (abs(p->getCharge()) != 1.0) {
+            continue;
+        }
+
+        auto tracks = p->getTracks();
```
tmadlener, line 248:
```suggestion
       const auto& tracks = p->getTracks();
```

Analysis/LeptonID/src/LeptonIDProcessor.cc
```diff
@@ -0,0 +1,356 @@
...
+        int nTracks = tracks.size();
+        if (nTracks != 1) {
+            continue;
+        }
+
+        auto clusters = p->getClusters();
```
tmadlener, line 254:
```suggestion
       const auto& clusters = p->getClusters();
```

tmadlener (approved):
Thanks for updating and incorporating feedback. I found a few unnecessary copies of `std::vector` still, due to `auto` not deducing references. I will commit them via the web interface.

Then we do the rest in the next PR(s).

Analysis/LeptonID/src/PIDutil.cc
```diff
@@ -0,0 +1,137 @@
+#include "PIDutil.h"
+#include <cmath>
+#include <set>
+
+#include <marlin/VerbosityLevels.h>
+
+#include <marlinutil/MarlinUtil.h>
+
+#include <EVENT/ReconstructedParticle.h>
+
```
tmadlener, line 10:
```suggestion
#include <UTIL/LCRelationNavigator.h>
```

tmadlener:
clicdp nightly CI is failing because it picks up the wrong LCIO. Let me try to fix that with another PR

tmadlener:
#115 should fix the issues with the clicdp nightlies based CI builds.

tmadlener:
@Zehvogel do you want to pick up iLCSoft/LCIO#171 still for this?

Zehvogel:
@tmadlener I don't have a strong preference. If you want I can make the changes now.

Zehvogel:
Looks like it will not be that easy to do... `getRelatedToMaxWeightAndObject` returns a `const LCObject*` but needs non-const `LCObject*` as input as the underlying functions expect that. And I want to call it again with the output of the first call as input... 

I had a quick look on how much of LCIO one would need to touch to have a more consistent use of `const` by changing `getRelatedToObjects` to take in a `const LCObject*`, but there is really no end in sight. 
We could also consider removing the `const` on the return value of `getRelatedToMaxWeightAndObject` etc. but I don't like that.

I can probably build a workaround using the From methods without touching LCIO again, but for now I would prefer to keep this as is and maybe change it later after someone has made LCIO `const`-ness a bit more consistent :)

tmadlener:
I will rebase this onto the latest master then, and merge after CI has passed (at least the ones that are currently working), unless you stop me in the next few hours.

PR #115, opened 2023-06-12
Fix linking order

tmadlener:
BEGINRELEASENOTES
- Bump the minimal required cmake version to 3.12
- Fix the order in which DD4hep and LCIO appear in the libraries to link against to make sure to not pick an inappropriate version from the underlying environment

ENDRELEASENOTES


tmadlener:
Merging this now, as it should fix the CI for #114. We can fix whatever needs changing in a follow up PR.

PR #117, opened 2023-06-19
Use digitized (not sim) SET hits for time-of-flight measurement

dudarboh:
BEGINRELEASENOTES

- In case of the option to measure time-of-flight at the SET: now using raw (digitized strips) SET hits instead of simulated hits. It doesn't change anything now, but could become trouble in the future when proper time digitization is added.

ENDRELEASENOTES

### DIFF
--- a/TimeOfFlight/src/TOFEstimators.cc
+++ b/TimeOfFlight/src/TOFEstimators.cc
@@ -137,14 +137,14 @@ void TOFEstimators::processEvent(EVENT::LCEvent * evt){
                     streamlog_out(WARNING)<<"Found no raw SET strip hits, but space point is built!? Writing TOF as 0."<<std::endl;
                 }
                 else if( rawObjects.size() == 1 ){
-                    streamlog_out(WARNING)<<"Found only one SET strip hit, but space point is built!? Writing TOF from a single strip."<<std::endl;
-                    TrackerHitPlane* rawHit = dynamic_cast <TrackerHitPlane*>(rawObjects[0]);
+                    streamlog_out(ERROR)<<"Found only one SET strip hit, but space point is built!? Writing TOF from a single strip."<<std::endl;
+                    auto* rawHit = dynamic_cast <TrackerHitPlane*>(rawObjects[0]);
                     if ( rawHit != nullptr ) timeOfFlight = RandGauss::shoot(rawHit->getTime(), _timeResolution);
                 }
                 else{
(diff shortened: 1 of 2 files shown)


TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -133,26 +130,26 @@ void TOFEstimators::processEvent(EVENT::LCEvent * evt){
...
+                const std::vector<LCObject*>& rawObjects = hitSET->getRawHits();
+                if ( rawObjects.empty() ){
+                    streamlog_out(WARNING)<<"Found no raw SET strip hits, but space point is built!? Writing TOF as 0."<<std::endl;
                 }
-                else if (simHitsSET.size() == 1){
-                    streamlog_out(WARNING)<<"Found only one SET strip hit! Writing TOF from a single strip."<<std::endl;
-                    SimTrackerHit* simHitSET = static_cast <SimTrackerHit*>(simHitsSET[0]);
-                    timeOfFlight = RandGauss::shoot(simHitSET->getTime(), _timeResolution);
+                else if( rawObjects.size() == 1 ){
+                    streamlog_out(WARNING)<<"Found only one SET strip hit, but space point is built!? Writing TOF from a single strip."<<std::endl;
```
tmadlener, line 140: How often do these happen?

TimeOfFlight/src/TOFEstimators.cc
```diff
@@ -133,26 +130,26 @@ void TOFEstimators::processEvent(EVENT::LCEvent * evt){
...
+                    TrackerHitPlane* rawHit = dynamic_cast <TrackerHitPlane*>(rawObjects[0]);
```
tmadlener, line 141:
```suggestion
                    auto* rawHit = dynamic_cast <TrackerHitPlane*>(rawObjects[0]);
```
Not really necessary to fix, but e.g. `clang-tidy` would flag this ad duplicating the type unnecessarily.

tmadlener (approved):
Sorry, completely missed this before. Effectively this can go in as is. Just one question about the frequency of one of the warnings (which was already there before), so probably rather infrequent.

TimeOfFlight/src/TOFEstimators.cc
dudarboh: This should never happen, in principle. It is mathematically impossible to reconstruct the space point with a single strip. It is here just in case I am wrong for some reason...

dudarboh:
Messed up commits a bit.. But now everything should be ok

PR #119, opened 2023-06-26
Add SLDCorrection to MarlinReco

yradkhorrami:
BEGINRELEASENOTES
- Merge SLDCorrection to MarlinReco
- First implementation of SLDCorrection: for the time being, done only on semi-leptonic decays of bottom hadrons in b-jets
ENDRELEASENOTES


tmadlener (commented):
Thank you for bringing this over to MarlinReco. I didn't have a too detailed look yet, so these are just some first rather general comments. I will have a closer look later and will then also have some more detailed comments.

It seems that your branch is currently slightly out of date with the master branch of MarlinReco. From the merge conflict it looks like it should be rather straight forward to fix, the main question here is whether I should do it or whether you want to do it.

As a general comment, there are many utility functions that could really use docstrings that very briefly describe what they are doing, but also which parameters are inputs and which are used to return things from the function. Currently it is a bit hard to figure these things out.

There are quite a few commented utility functions as well. They seem to be from development, but judging from the names some of them might still be useful for analysis(?). I think here we should decide which to keep (and uncomment) and which to remove entirely.

There are also a few typedefs that are repeated over many header files. For consistency (and maintainability) I think it would be a good idea to create a header, e.g. `SLDCorrectionDefinitions.h` and put all of them there. Additionally, these should be capitalized otherwise it becomes very confusing to disambiguate between variable names and typedefs.

tmadlener:
superseeded by #122

PR #121, opened 2023-07-03
LeptonID: remove notebook output

Zehvogel:
The jsroot plots were not showing up correctly on github anyway...

BEGINRELEASENOTES
- LeptonID: remove notebook output

ENDRELEASENOTES


tmadlener:
Could it be turned into a python script that spits out the plots easily? Could be useful for redoing the training or checking the plots.

Zehvogel:
Would be easy to auto convert to a python script. Don't know how easy it would be to go back from that to a notebook again for further interactive studies of possible input variables...

jmcarcell:
It's not a good idea to keep complex analysis code in notebooks, you lose a lot on version control, reproducibility, reusability, etc

tmadlener:
@Zehvogel technically you can always `from script import *` in a notebook. In the end, I do agree with Juan here; A notebook is not a place to store any code that is meant to be run repeatedly. If you want the converted python script here, I would wait with merging this one now. I will merge this in the early afternoon and make a new tag (in whatever state this is at that point).

Zehvogel:
Yes, I agree that repeated runs are not the thing to do with a notebook and of course one can always include all the things :)
The purpose of the notebook is to quickly plot comparisons between electrons, muons and charged hadrons distributions for new possible input variables... Of course also possible from a script.

I would be happy if you merge this as is :)

tmadlener (approved):

PR #122, opened 2023-07-03
Add SldCorrection

yradkhorrami:
BEGINRELEASENOTES
- Merge SLDCorrection to MarlinReco
- First implementation of SLDCorrection: for the time being, done only on semi-leptonic decays of bottom hadrons in b-jets
ENDRELEASENOTES


Analysis/SLDCorrection/src/FindParticle.cc
```diff
@@ -0,0 +1,238 @@
...
+	weightMCPtoPFO = 0.0;
+	double maxweightPFOtoMCP = 0.;
+	double maxweightMCPtoPFO = 0.;
+	int iPFOtoMCPmax = -1;
+	int iMCPtoPFOmax = -1;
+	for ( unsigned int i_pfo = 0; i_pfo < PFOvec.size(); i_pfo++ )
+	{
+		double pfo_weight = 0.0;
+		double trackWeight = ( int( PFOweightvec.at( i_pfo ) ) % 10000 ) / 1000.0;
+		double clusterWeight = ( int( PFOweightvec.at( i_pfo ) ) / 10000 ) / 1000.0;
```
tmadlener, line 146: You could use the [new functionality in MarlinUtil](https://github.com/iLCSoft/MarlinUtil/blob/13d75c07698d28564aa6366c61dc6c0b20d58769/source/include/MarlinUtil.h#L64-L80) for this.

Analysis/SLDCorrection/include/SLDCorrection.h
```diff
@@ -0,0 +1,598 @@
...
+	IntVector				m_recoSolutionSign{};
+	IntVector				m_nChargedParticlesInPrimVertex{};
+	IntVector				m_nChargedParticlesNotInPrimVertex{};
+	FloatVector				m_weightPFOtoMCP_Lepton{};
+	FloatVector				m_weightMCPtoPFO_Lepton{};
+	FloatVector				m_weightPFOtoMCP_Neutral{};
+	FloatVector				m_weightMCPtoPFO_Neutral{};
+	FloatVector				m_weightPFOtoMCP_Charged{};
+	FloatVector				m_weightMCPtoPFO_Charged{};
+	TH1I					*h_SLDStatus{};
```
tmadlener, line 558: All of these histograms seem to be leaked (or at least I can't see where they are cleaned up). I would suggest making them `std::unique_ptr` instead of raw pointers and then use `std::make_unique` instead of `new` to create them. This way they are automatically cleaned up. E.g.

```cpp
// member declaration in the header
std::unique_ptr<TH1I> h_SLDStatus{nullptr};

// initialization in the source file
h_SLDStatus = std::make_unique<TH1I>("SLDStatus" , ";" , 7 , 0 , 7);
```

tmadlener (commented):
Hi Yasser,

thanks for cleaning up the code a bit. I have had another look and there are a couple of things that I think would still be good to address:

- Many (almost all) of the functions in the header files could really benefit from having a short description of what they do (and also what the meaning of their return values and parameters is). Since the function names are already quite descriptive a one sentence summary is probably enough for most of them. For some of them you will probably have to add a brief description of the parameters as well.
- Some of the functions that are defined, e.g. in `AssignParticlestoSLD.h` take rather "heavy" objects" by value (e.g. a `vector` or a `LCRelationNavigator`), which is almost never what you want, because this creates a copy of the input every time you call it, instead:
 - If the input is not modified by a function pass it as a `const` reference (e.g. `const vector<XYZ>&` or `const LCRelationNavigator&`).
 - If you want to modify such an object inside a function such that the changes are visible outside, pass it as a normal reference (e.g. `vector<XYZ>&`, `LCRelationNavigator&`).
- The histograms in the `SLDCorrection` processor are leaked at the moment. I have added a possible fix in a comment below. I am not entirely sure if you want these histograms in the output file or not, but using `unique_ptr` might interfere with that.
- For all the variables you write into a tree you currently introduce separate variables and then individually make a branch for each of them. It is probably not worth rewriting this now, but this could be solved much more easily with a `map` or an `unordered_map` and a loop.

Analysis/SLDCorrection/include/SLDCorrection.h
```diff
@@ -0,0 +1,598 @@
...
+
+	int					m_nRun;
+	int					m_nEvt;
+	int					m_nRunSum;
+	int					m_nEvtSum;
+	double					m_Bfield;
+	double					c;
+	double					mm2m;
+	double					eV2GeV;
+	double					eB;
```
tmadlener, line 165: These seem to be unused in the source code (other than their initialization).

tmadlener:
superseeded by #123

PR #132, opened 2024-05-07
Make the TrueJet processor use the PIDHandler to set the ParticleIDs 

tmadlener:
BEGINRELEASENOTES
- Make the `TrueJet` processor use the `PIDHandler` to set the `ParticleIDs` for the different objects it creates. This sets the necessary metadata that is required, e.g. for the conversion to EDM4hep. 

ENDRELEASENOTES

Fixes key4hep/k4EDM4hep2LcioConv#62

@Zehvogel can you check whether this fixes things on your end? At least from a technical level, the algorithm names might still need some tuning, but I would also be interested to see which ones are actually hit / converted.

### DIFF
--- a/Analysis/TrueJet/src/TrueJet.cc
+++ b/Analysis/TrueJet/src/TrueJet.cc
@@ -705,10 +705,10 @@ void TrueJet::processEvent( LCEvent * event ) {
 
       LCCollectionVec* fafpf_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
       auto fafpfPidHandler = UTIL::PIDHandler(fafpf_vec);
-      auto fafpf_mainPidId = fafpfPidHandler.addAlgorithm("TrueType_fafpf", {});
-      std::array<int, 25> fafpf_pidIds{};
+      auto fafpf_mainPidId = fafpfPidHandler.addAlgorithm("TrueJet_fafpf", {});
+      std::array<int, 2> fafpf_pidIds{};
       for (size_t i = 0; i < fafpf_pidIds.size(); ++i) {
-          fafpf_pidIds[i] = fafpfPidHandler.addAlgorithm("TrutType_fafpf_jet_" + std::to_string(i), {});
+          fafpf_pidIds[i] = fafpfPidHandler.addAlgorithm("TrueJet_fafpf_jet_" + std::to_string(i), {});
       }
 
       LCRelationNavigator FinalColourNeutral_Nav(LCIO::RECONSTRUCTEDPARTICLE , LCIO::RECONSTRUCTEDPARTICLE ) ;
@@ -832,7 +832,7 @@ void TrueJet::processEvent( LCEvent * event ) {
                                       0, fafpf_mainPidId, {});
 
         for ( int j_jet_end=1 ; j_jet_end<=jets_end[0][k_dj_end] ; j_jet_end++ ) {
-          fafpfPidHandler.setParticleID(fafpf, type[jets_end[j_jet_end][k_dj_end]], pdg[j_jet_end], 0, fafpf_pidIds[j_jet_end], {});
+          fafpfPidHandler.setParticleID(fafpf, type[jets_end[j_jet_end][k_dj_end]], pdg[j_jet_end], 0, fafpf_pidIds[j_jet_end - 1], {});
 
           if ( elementon[jets_end[j_jet_end][k_dj_end]] > 0 ) {
             FinalElementon_Nav.addRelation(  jet_vec->getElementAt(jets_end[j_jet_end][k_dj_end]-1) , 
@@ -849,10 +849,10 @@ void TrueJet::processEvent( LCEvent * event ) {
 
       LCCollectionVec* fafpi_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
       auto fafpiPidHandler = UTIL::PIDHandler(fafpi_vec);
-      auto fafpi_mainPidId = fafpiPidHandler.addAlgorithm("TrueType_fafpi", {});
+      auto fafpi_mainPidId = fafpiPidHandler.addAlgorithm("TrueJet_fafpi", {});
       std::array<int, 25> fafpi_pidIds{};
       for (size_t i = 0; i < fafpi_pidIds.size(); ++i) {
-          fafpi_pidIds[i] = fafpiPidHandler.addAlgorithm("TrutType_fafpi_jet_" + std::to_string(i), {});
+          fafpi_pidIds[i] = fafpiPidHandler.addAlgorithm("TrueJet_fafpi_jet_" + std::to_string(i), {});
       }
 
       LCRelationNavigator InitialElementon_Nav(LCIO::RECONSTRUCTEDPARTICLE , LCIO::MCPARTICLE ) ;
@@ -981,7 +981,7 @@ void TrueJet::processEvent( LCEvent * event ) {
         fafpiPidHandler.setParticleID(fafpi, type[jets_begin[1][k_dj_begin]] % 100, pdg[0], 0, fafpi_mainPidId, {});
 
         for(int j_jet_begin=1 ; j_jet_begin<=jets_begin[0][k_dj_begin] ; j_jet_begin++) {
-          fafpiPidHandler.setParticleID(fafpi, type[jets_begin[j_jet_begin][k_dj_begin]], pdg[j_jet_begin], 0, fafpi_pidIds[j_jet_begin], {});
+          fafpiPidHandler.setParticleID(fafpi, type[jets_begin[j_jet_begin][k_dj_begin]], pdg[j_jet_begin], 0, fafpi_pidIds[j_jet_begin - 1], {});
           fafpi->addParticle(dynamic_cast<ReconstructedParticle*>(jet_vec->getElementAt(jets_begin[j_jet_begin][k_dj_begin]-1)));
         }
         fafpi_vec->addElement(fafpi);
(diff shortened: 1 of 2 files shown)


Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -698,6 +704,12 @@ void TrueJet::processEvent( LCEvent * event ) {
         // post-PS part
 
       LCCollectionVec* fafpf_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
+      auto fafpfPidHandler = UTIL::PIDHandler(fafpf_vec);
+      auto fafpf_mainPidId = fafpfPidHandler.addAlgorithm("TrueType_fafpf", {});
+      std::array<int, 25> fafpf_pidIds{};
+      for (size_t i = 0; i < fafpf_pidIds.size(); ++i) {
+          fafpf_pidIds[i] = fafpfPidHandler.addAlgorithm("TrutType_fafpf_jet_" + std::to_string(i), {});
```
tmadlener, line 711:
```suggestion
          fafpf_pidIds[i] = fafpfPidHandler.addAlgorithm("TrueJet_fafpf_jet_" + std::to_string(i), {});
```

Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -698,6 +704,12 @@ void TrueJet::processEvent( LCEvent * event ) {
         // post-PS part
 
       LCCollectionVec* fafpf_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
+      auto fafpfPidHandler = UTIL::PIDHandler(fafpf_vec);
+      auto fafpf_mainPidId = fafpfPidHandler.addAlgorithm("TrueType_fafpf", {});
```
tmadlener, line 708:
```suggestion
      auto fafpf_mainPidId = fafpfPidHandler.addAlgorithm("TrueJet_fafpf", {});
```

Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -837,6 +848,12 @@ void TrueJet::processEvent( LCEvent * event ) {
         // pre-PS part
 
       LCCollectionVec* fafpi_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
+      auto fafpiPidHandler = UTIL::PIDHandler(fafpi_vec);
+      auto fafpi_mainPidId = fafpiPidHandler.addAlgorithm("TrueType_fafpi", {});
```
tmadlener, line 852:
```suggestion
      auto fafpi_mainPidId = fafpiPidHandler.addAlgorithm("TrueJet_fafpi", {});
```

Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -837,6 +848,12 @@ void TrueJet::processEvent( LCEvent * event ) {
...
+      std::array<int, 25> fafpi_pidIds{};
+      for (size_t i = 0; i < fafpi_pidIds.size(); ++i) {
+          fafpi_pidIds[i] = fafpiPidHandler.addAlgorithm("TrutType_fafpi_jet_" + std::to_string(i), {});
```
tmadlener, line 855:
```suggestion
          fafpi_pidIds[i] = fafpiPidHandler.addAlgorithm("TrueJet_fafpi_jet_" + std::to_string(i), {});
```

Zehvogel:
Yes I will test it. It would probably also be good if someone who uses the MarlinUtil TrueJet parser can test if that still works with this change. Maybe someone at DESY? :)

tmadlener:
Thanks. In principle nothing should have changed other than that we are now writing some additional metadata as well. I will check if that is actually true.

Zehvogel:
This fixes running TrueJet for me. I still need to adapt my RDataframe code to work on the new direction though.
However I compared the `podio-dump` output after this change and before reversal of the PID collection and it looks like everything is correct. Depending on what kind of events one uses a lot of the PIDs will stay empty.

As far as I can tell `*_PID_TrueJet_*_jet_n` is filled for the n-th jet belonging to the initial/final colour neutral. I have no idea how this will ever reach values much greater than two... maybe @MikaelBerggren can comment if we really need 25? Also the 0'th one is _never_ filled? The files become really crowded at the moment with all of them in there...

I attached the above-mentioned `podio-dump` outputs if anyone wants to take a look :)

[truejet_oldpid.txt](https://github.com/iLCSoft/MarlinReco/files/15250763/truejet_oldpid.txt)
[truejet_newpid.txt](https://github.com/iLCSoft/MarlinReco/files/15250761/truejet_newpid.txt)

Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -698,6 +704,12 @@ void TrueJet::processEvent( LCEvent * event ) {
         // post-PS part
 
       LCCollectionVec* fafpf_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
+      auto fafpfPidHandler = UTIL::PIDHandler(fafpf_vec);
+      auto fafpf_mainPidId = fafpfPidHandler.addAlgorithm("TrueJet_fafpf", {});
+      std::array<int, 25> fafpf_pidIds{};
+      for (size_t i = 0; i < fafpf_pidIds.size(); ++i) {
+          fafpf_pidIds[i] = fafpfPidHandler.addAlgorithm("TrueJet_fafpf_jet_" + std::to_string(i), {});
```
tmadlener, line 711:
```suggestion
          fafpf_pidIds[i] = fafpfPidHandler.addAlgorithm("TrueJet_fafpf_jet_" + std::to_string(i + 1), {});
```
> Also the 0'th one is never filled? The files become really crowded at the moment with all of them in there...

Zehvogel: doesn't this do the opposite? I.e. it just renames 0 to 1 but it will still never be filled?

tmadlener: It doesn't fix that, but I just remembered I wanted to do this, because previously the one that goes currently into the "main" PID went into `0`.

tmadlener:
> Also the 0'th one is never filled? The files become really crowded at the moment with all of them in there...

I am not sure if this can even be checked easily, but were these filled before?

Zehvogel:
> > Also the 0'th one is never filled? The files become really crowded at the moment with all of them in there...
> 
> I am not sure if this can even be checked easily, but were these filled before?

I think that the 0'th one was not filled was expected if you start the counting of the jets per ICN by 1. There is no ICN/FCN with no truejet belonging to it. So that one seems to be simply left over.

For the values > 2 I am not so sure. I think 3 and 4 are used relatively often (as soon as you have gluon-splitting etc.).

Leaving my area of expertise even more I would guess that a _final_ colour neutral would by definition always result in two (true)jets while an _initial_ colour neutral can have $2n$ (true)jets where $n$ is the number of _final_ colour neutrals it hadronises into.

So maybe we can at least reduce the amount of `fafpf` PIDs to two (three including the main one)?

tmadlener:
After merging #134, we have almost halfed the number of ParticleID objects that will be visible once this has been converted to EDM4hep. For the initial state side there is no generally applicable way to do this. The number 25 has been chosen to work at ILD up to 1 TeV. At lower energies this might be smaller. I think the easiest way of dealing with this might be to simply drop the "superfluous" collections via the `KeepDropSwitch` and configuration (untested) along the lines of 
```python
# ...
"drop *_PID_TrueJet_fafpi_jet1?",
"drop *_PID_TrueJet_fafpi_jet2?",
# and then maybe some single digits as well if you still want to get rid of more
# or alternatively drop all of them and keep only the ones you want, might be the same number of lines
```

Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -837,6 +848,12 @@ void TrueJet::processEvent( LCEvent * event ) {
         // pre-PS part
 
       LCCollectionVec* fafpi_vec = new LCCollectionVec(LCIO::RECONSTRUCTEDPARTICLE ) ;
+      auto fafpiPidHandler = UTIL::PIDHandler(fafpi_vec);
+      auto fafpi_mainPidId = fafpiPidHandler.addAlgorithm("TrueJet_fafpi", {});
+      std::array<int, 25> fafpi_pidIds{};
```
Zehvogel, line 853: I know the 25 is just the result of correctly making the 0'th pid the mainPid, but I wonder if the 26 before was correct? I would have thought that we would expect an even number here (after removal of the mainPid).

Analysis/TrueJet/src/TrueJet.cc
```diff
@@ -837,6 +848,12 @@ void TrueJet::processEvent( LCEvent * event ) {
...
+      for (size_t i = 0; i < fafpi_pidIds.size(); ++i) {
+          fafpi_pidIds[i] = fafpiPidHandler.addAlgorithm("TrueJet_fafpi_jet_" + std::to_string(i), {});
```
Zehvogel, line 855: can we format this to left-pad with 0 to have nicer sorting in `podio-dump` etc.?

tmadlener: So basically, `TrueJet_fafpi_jet_01` for example?

Zehvogel: yes

Zehvogel:
While you are in there anyway, do you mind setting
https://github.com/iLCSoft/MarlinReco/blob/8c3f3b614395d31dbaf85b09491d1307226e5d26/Analysis/TrueJet/src/TrueJet.cc#L157
to something like `MESSAGE`?

Analysis/TrueJet/src/TrueJet.cc
tmadlener: done. Should work until someone wants more than 99 of these ;)

PR #134, opened 2024-06-10
TrueJet: Reduce the size of the ParticleID vector for final fermion anti fermion pair. Update doc and examples.

MikaelBerggren:
BEGINRELEASENOTES

Reduce the size of the ParticleID vector for the final fermion-antifermion pair, since

for this case, there can only be one pair. This to avoid cluttering of empty collections after transition to the EDM4HEP world. At the same time, the documentation and example steerings in the examples subdirectory have been updated. mainly for the move of TrueJet_Parser from here to MarlinUtil, but also spell-checking etc. 

ENDRELEASENOTES


Analysis/TrueJet/examples/truejet_CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
+########################################################
```
tmadlener, line 1: I assume this is the `CMakeLists.txt` file that allows a standalone compilation of the `TrueJet` module? In that case, we could move this up to `Analysis/TrueJet`. It doesn't interfere there with the normal build process for MarlinReco, and it would make building TrueJet in a standalone fashion a bit easier, since there is no renaming of the `trujet_CMakeLists.txt` file necessary.

tmadlener (approved):
Merging this to let #132 move forward

PR #136, opened 2024-07-15
Fix small bug in WW categorization

ueinhaus:
BEGINRELEASENOTES
Bug: PFOs were ignored if their MC PDG was not among signal or background PDGs. This is of minor effect, since by default all detector-stable charged particles are considered signal or background, but could lead to MC info leaking into reconstructed values in case of unintended usage.
Solution: This effect now requires training mode to be ON, which is exclusive with inference mode.
ENDRELEASENOTES


tmadlener (commented):
IIUC, the only "real" change is moving existing code into another if here, right? (And adding one more printout that wasn't there before).

Issue #140, opened 2024-09-24
New warnings with gcc14


tmadlener:
See https://github.com/iLCSoft/MarlinReco/actions/runs/11013059693/job/30580558907#step:3:491

PR #141, opened 2024-09-30
Tof analysis

dudarboh:
BEGINRELEASENOTES

- Adding time-of-flight particle identification analysis codes from Bohdan Dudar's PhD studies.

ENDRELEASENOTES


dudarboh:
Sorry, I only noticed #140 ...
Do we want to keep the warnings for now?

tmadlener:
I can lift the commit with the "external warnings" into a separate PR. Thanks for taking care of it, I haven't gotten around to it yet.

tmadlener:
Warnings of #140 have been fixed in #142. A simple rebase here should do the rest.

tmadlener:
I think the analysis and plotting code would be better suited for [ILDPerformance](https://github.com/ilcsoft/ildperformance), where a `TimeOfFlight` subfolder already exists. Is it somewhat easy to split that off from here?

dudarboh:
I liked the idea that one can find all the codes related to my study in one place given only one link.
But I see that it is a bit inconsistent with other parts of the software...

I can move the analysis folder with all Python scripts to the ILDPerformance with a note that they are meant to be used with a ROOT file produced from this processor in MarlinReco.
Do you have any opinion on how to organise the old code from ILDPerformance/TimeOfFlight? Should everything else be put into the `old` folder? Maybe you have better ideas for a better name. Or is it so obsolete that it can be safely removed?

`./src/BohdanDrawing` is meant to be used inside the Marlin processor for debugging, so I would say it is better to stay here as I do not think it is very useful separately... Creating a separate processor just for these functions only is unnecessary, imho.

PR #143, opened 2024-11-05
Bugfixes in SLDCorrection

nVentis:
BEGINRELEASENOTES
- fixed segfaults in SLDCorrection:
- affects all versions: implemented checks if assumed elements exist
- affects recent key4hep versions: delete of histograms crashing the end method

ENDRELEASENOTES


Analysis/SLDCorrection/src/FlightDirection.cc
```diff
@@ -33,6 +33,12 @@ int getRecoFlightDirection( const RecoParticle &linkedRecoLepton , TVector3 &rec
 	daughterHadronFlightDistance = 0.0;
 	daughterHadronFlightDirection = TVector3( 0.0 , 0.0 , 0.0 );
 	sldVertexPosition.clear();
+	if ( linkedRecoLepton->getTracks().size() == 0 )
```
tmadlener, line 36:
```suggestion
	if ( linkedRecoLepton->getTracks().empty() )
```
This is the recommended (and more readable) way for checking whether a vector has no elements.

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -801,7 +801,7 @@ void SLDCorrection::init()
 		m_pTTree1->Branch( "DSVDistanceFromPV" , &m_DSVDistanceFromPV );
 		m_pTTree1->Branch( "Lepton3DImpactParameter" , &m_Lepton3DImpactParameter );
 		m_pTTree1->Branch( "OtherParticle3DImpactParameter" , &m_OtherParticle3DImpactParameter );
-		h_SLDStatus = new TH1I( "SLDStatus" , ";" , 7 , 0 , 7 );
+		h_SLDStatus = new TH1I( "SLDStatus" , ";" , 8 , -1 , 7 );
```
tmadlener, line 804: I think the number of bins should change here as well? Or was it incorrect before? I suppose the main reason for doing this is to have access to things that haven't been classified into any of the pre-assigned categories? However, the previous version had that already if you use the under-/overflow bins.

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -3536,39 +3547,7 @@ void SLDCorrection::end()
...
-		delete h_SLDecayModeC;
-		delete h_SLDecayOrder;
-		delete h_foundVertex;
-		delete h_secondaryVertex;
-		delete h_parentHadronCharge;
-		delete h_MCPTracks;
-		delete h_MCPTracks_Eweighted;
-		delete h_MCPTracks_Ptweighted;
-		delete h_FlightDirectionError;
-		delete h_distRecoLeptonToDownStreamVertex;
```
tmadlener, line 3571: Why are these removed?

tmadlener (commented):
Thanks, especially also for documenting some parts. I have a few comments / questions, see below.

Analysis/SLDCorrection/src/FlightDirection.cc
nVentis: Thanks! Will include this suggestion here and everywhere where the same check occurs.

Analysis/SLDCorrection/src/SLDCorrection.cc
nVentis: Before, the values were before 0/7, now -1/7 (throughout the code, floats are inserted, i.e. 0.5 -> 1. I didn't change this). According to [here](https://root.cern.ch/doc/master/classTH1.html) the third argument is the bin count, so it should be fine?

Analysis/SLDCorrection/src/SLDCorrection.cc
nVentis: In older key4hep releases (e.g. 2024-04-10), there were no problems with the histograms, but in a more recent stack (e.g. 2024-10-XX), I had segfaults related to the ::end method of the SLD Correction. This affected the output of other processors as well. Maybe there was a change in how ROOT keeps track of stuff in between?
When I removed the delete statements, it was working fine.

Analysis/SLDCorrection/src/SLDCorrection.cc
tmadlener: Ah right, thanks for the clear up. I was once more confusing the order of arguments in the `TH1I` constructor. But as far as I can see there is no actual need for a negative bin with value `-1`? `h_SLDStatus` is only filled with the values `-0.5`, `0.5` and `1.5` at the moment.

tmadlener:
Apologies, this seems to have dropped from my list somehow. Is this ready to be merged, or are is there more work to come?

tmadlener:
Is there anything missing from this or can this be merged?

tmadlener:
The builds that I would expect to succeed are succeeding. The diff of `TrueJet` is huge and it's not quite clear to me whether it is large because a lot of things changed or whether it is large because just enough pieces move around to create a huge diff despite rather few things changing overall.

Can you maybe give a short summary on what actually changes on a high level?

nVentis:
Sorry @tmadlener I did not actually mean to push to this PR. I'll clean this up and create a separate PR for the fixes to TrueJet (the idea there was to catch cases which would lead to segmentation fault, in which case TrueJet would skip the current event).

Victor-Schwan:
What is the status here? Soon, I wanna use the `SLDCorrection` as well and would appreciate if this PR is merged by then :)

nVentis:
I've polished the PR and included all the changes to the SLDCorrection processor that we're using for the ZHH analysis. Ready for merging

Victor-Schwan:
@tmadlener what is the status here?

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -300,9 +291,10 @@ void SLDCorrection::init() {
   DDMarlinCED::init(this);
 
   if (m_fillRootTree) {
-    m_pTFile = new TFile(m_rootFile.c_str(), "recreate");
-    m_pTTree1 = new TTree("SLDCorrection", "SLDCorrection");
-    m_pTTree1->SetDirectory(m_pTFile);
+    if (m_rootFile.size()) {
```
tmadlener:
```suggestion
    if (!m_rootFile.empty()) {
```
Recommended way of checking for a non-empty string.

Analysis/SLDCorrection/src/FlightDirection.cc
```diff
@@ -48,6 +48,12 @@ int getRecoFlightDirection(const RecoParticle& linkedRecoLepton, TVector3& recoF
   daughterHadronFlightDistance = 0.0;
   daughterHadronFlightDirection = TVector3(0.0, 0.0, 0.0);
   sldVertexPosition.clear();
+  if (linkedRecoLepton->getTracks().size() == 0) {
```
tmadlener:
```suggestion
  if (linkedRecoLepton->getTracks().empty()) {
```
Recommended way of checking for empty vectors.

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -3672,7 +3678,10 @@ void SLDCorrection::check(EVENT::LCEvent* pLCEvent) {
 
 void SLDCorrection::end() {
   if (m_fillRootTree) {
-    m_pTFile->cd();
+    if (m_rootFile.size()) {
```
tmadlener:
```suggestion
    if (!m_rootFile.empty()) {
```

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -3739,6 +3748,10 @@ void SLDCorrection::end() {
     delete h_MCPTracks_Ptweighted;
     delete h_FlightDirectionError;
     delete h_distRecoLeptonToDownStreamVertex;
-    delete m_pTFile;
+
+    if (m_pTFile != NULL) {
```
tmadlener:
```suggestion
    if (m_pTFile) {
```

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -1652,6 +1646,20 @@ void SLDCorrection::doSLDCorrection(EVENT::LCEvent* pLCEvent, const MCP& SLDLept
...
   LCCollection* primaryVertexCollection = pLCEvent->getCollection(m_inputPrimaryVertex);
+
+  MCP parentHadron = SLDLepton->getParents()[parentHadronIDx];
+
+  int SLDStatus = -999;
+  if (primaryVertexCollection->getNumberOfElements() == 0) {
+    SLDStatus = 0;
+    streamlog_out(WARNING) << "	||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||" << std::endl;
+    streamlog_out(WARNING) << "	||||||||||||||||||||| PrimaryVertex is not found |||||||||||||||||||||" << std::endl;
+    streamlog_out(WARNING) << "	||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||" << std::endl;
```
tmadlener: Given that this is a "regular" entry into the `h_SLDStatus` monitoring histogram, is this really a warning? Or could it be demoted to info / message? How often does this happen?

Analysis/SLDCorrection/src/SLDCorrection.cc
```diff
@@ -1920,11 +1928,10 @@ void SLDCorrection::doSLDCorrection(EVENT::LCEvent* pLCEvent, const MCP& SLDLept
   ////////////////////////////////////////////////////////////////////////////////
   ////////////////////////////////////////////////////////////////////////////////
 
-  int SLDStatus = -999;
-  if (linkedRecoLepton == NULL) {
+  if (linkedRecoLepton == NULL || linkedRecoLepton->getTracks().size() == 0) {
```
tmadlener:
```suggestion
  if (linkedRecoLepton == NULL || linkedRecoLepton->getTracks().empty()) {
```

tmadlener (commented):
I have rebased this and fixed the merge conflict. Current CI failures are unrelated to the changes. There are a few minor stylistic fixes, I will put them into a commit. The one open question is the log level for the *no primary vertex found* case. Otherwise this is good to go also from my side.

Could you summarize in the top level comment how the seg-fault was fixed, so that there is a bit more context to the changes in case someone comes back to this in the future?

Issue #144, opened 2025-01-07
Photon Energy Correction is not propagated to the cluster


dudarboh:
[This processor](https://github.com/iLCSoft/MarlinReco/blob/dcaff65b1487c5ecfa8691eb57f3575f5902a636/Analysis/PhotonEnergyCorrect/src/photonCorrectionProcessor.cc) updates the energy of the reconstructed photons on the PFO level, but does not update it's linked cluster energy.

Thus, for all photons the energies between pfo and cluster are inconsistent
```cpp
if (pfo->getPDG() == 22)
    bool isConsistent = (pfo->getEnergy() == pfo->getClusters()[0]->getEnergy()) //false
```

In the standard production for some detector geometries, the processor is switched on. For some off.
It's not clear if this decision was intentional or if some detector models were not used for too long and calibration files are obsolete now.
For most widely used, e.g. `ILD_l5_o1_v02` it is on by default in the calibration file, e.g. [here](https://github.com/iLCSoft/ILDConfig/blob/19c63b1ca6d2038a0d640150575365cf0606b31c/StandardConfig/production/Calibration/Calibration_ILD_l5_o1_v02.xml#L26)

tmadlener:
I don't think this is necessarily an issue or at least it's probably at least partially also a question of "reconstruction philosophy". After all the correction is something you apply at the PFO level once you have identified it to be a photon. There is no inherent reason to propagate that correction back to the underlying cluster because that might lead to several potential issues:
- If you want to re-run Pandora the inputs now have changed and the results of Pandora now depend on whether you have run it before or not.
- How would you apply the correction if the photon is reconstructed from split clusters?
- How far back would you propagate the corrections? Clusters are made from hits, etc...

What is the reason for making this comparison in the first place?

dudarboh:
Hi, @danieljeans, thanks for the explanation.

Okay, if cluster and PFO energies are intentionally different, then there is no issue indeed. 
Maybe I confused myself because that was the first time I encountered different cluster/PFO energies.
Learned something new, I was always assuming they were the same.
I will close the issue.

PR #147, opened 2025-05-01
code to calculate photon PFO corrections

danieljeans:
BEGINRELEASENOTES
- Add code needed to obtain the photon energy and direction corrections which are used in the photonCorrectionProcessor (as promised in https://github.com/iLCSoft/MarlinReco/issues/144#issuecomment-2576491059 )
ENDRELEASENOTES

### DIFF
--- a/Analysis/PhotonEnergyCorrect/scripts/README.md
+++ b/Analysis/PhotonEnergyCorrect/scripts/README.md
@@ -2,22 +2,38 @@
 
 Daniel Jeans, 2022-2025
 
-calculate corrections to energy and direction of photon-like PFOs.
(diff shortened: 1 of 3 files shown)


tmadlener (commented):
Thanks for adding this. Is there an easy way to put a usage example into the README? I think that would be really helpful in the future for quickly getting things to run. It looks a bit like the scripts expect a certain directory structure to be present (or they create one?), so just some example invocations would be really nice to have.

danieljeans:
Thanks @tmadlener I expanded the README to explain in more detail how the user should proceed.

tmadlener (approved):
Thanks.

PR #155, opened 2025-12-16
Fix cone energy self-counting in IsolatedLeptonFinderProcessor

yyzhang666:
BEGINRELEASENOTES
- Fixes several related issues caused by comparing copied reconstructed particles with originals by pointer identity
- Avoids self-counting in isolation cone energy
- Ensures jet-based isolation uses original PFOs
- Completes copying of tracks and clusters when duplicating PFOs

ENDRELEASENOTES

### DIFF
--- a/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
+++ b/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
@@ -167,7 +167,21 @@ protected:
   /** If set to true, uses Pandora particle IDs */
   bool _usePandoraIDs = false;
 
-  /* DEBUG: Map from copied PFO to original PFO */
+ /**
+ * Map from internally copied PFOs to their original input PFOs.
+ *
+ * The IsolatedLeptonFinder creates internal copies of ReconstructedParticles
+ * (e.g. for dressing leptons or producing modified output collections).
+ * However, isolation quantities such as the cone energy are defined with
+ * respect to the original input PFO collection.
+ *
+ * Since the copied PFOs are distinct objects, pointer comparisons would
+ * otherwise fail (e.g. when excluding the lepton itself from the cone),
+ * leading to incorrect self-counting in the isolation energy.
+ *
+ * This map is therefore used to recover the association to the original
+ * PFO when computing isolation-related quantities.
+ */
   std::map<ReconstructedParticle*, ReconstructedParticle*> _copy2orig;
 };
 
--- a/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
+++ b/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
@@ -36,6 +36,11 @@ public:
   virtual void processEvent(LCEvent* evt);
   virtual void end();
 
+private:
+
+  /// Return the original PFO if this particle is a copy, otherwise return itself
+  ReconstructedParticle* findOriginal(ReconstructedParticle* pfo) const;
+
 protected:
   /** Returns true if pfo is a lepton */
   bool IsGoodLepton(ReconstructedParticle* pfo);
--- a/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
+++ b/Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
@@ -18,10 +18,10 @@
 #include <lcio.h>
 #include <marlin/Processor.h>
 
+#include "IMPL/ReconstructedParticleImpl.h"
 #include <EVENT/MCParticle.h>
 #include <EVENT/ReconstructedParticle.h>
 #include <UTIL/LCRelationNavigator.h>
-#include "IMPL/ReconstructedParticleImpl.h"
 
 class IsolatedLeptonFinderProcessor : public marlin::Processor {
 public:
@@ -29,7 +29,7 @@ public:
 
   IsolatedLeptonFinderProcessor();
 
-  IsolatedLeptonFinderProcessor(const IsolatedLeptonFinderProcessor&)            = delete;
+  IsolatedLeptonFinderProcessor(const IsolatedLeptonFinderProcessor&) = delete;
   IsolatedLeptonFinderProcessor& operator=(const IsolatedLeptonFinderProcessor&) = delete;
 
   virtual void init();
@@ -37,7 +37,6 @@ public:
   virtual void end();
 
 private:
-
   /// Return the original PFO if this particle is a copy, otherwise return itself
   ReconstructedParticle* findOriginal(ReconstructedParticle* pfo) const;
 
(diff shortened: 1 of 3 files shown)


tmadlener:
Thanks a lot for putting this together. A few comments / requests.
- This looks like it includes the changes from #143. Is that correct?
- There are quite a few merge conflicts in this PR. It might be that your work. This is almost certainly because it starts off quite a bit back from the master branch. This should probably be fixable by formatting the file accordingly, e.g. via `clang-format -i -style=file Analysis/*/*/*.{cc,h}`
- Please keep the original block from the template and fill the release notes in between them, as we use these tags to generate release notes for all PRs automatically (i.e. the `BEGINRELEASENOTES ... ENDRELEASENOTES`).

yyzhang666:
Apology that I'm new to all git stuff. 
>     * This looks like it includes the changes from [Bugfixes in SLDCorrection #143](https://github.com/iLCSoft/MarlinReco/pull/143). Is that correct?
>  I work from Bryan's repository so it may contains this #143 . I have rebased it to the current version officially here.

>     * There are quite a few merge conflicts in this PR. It might be that your work. This is almost certainly because it starts off quite a bit back from the master branch. This should probably be fixable by formatting the file accordingly, e.g. via `clang-format -i -style=file Analysis/*/*/*.{cc,h}`
> Now I only formatted the IsolatedLeptonFinder,  seems still some conflicts, I manually resolved them in Github here. But now sure if it is good, should I format all repository?

>     * Please keep the original block from the template and fill the release notes in between them, as we use these tags to generate release notes for all PRs automatically (i.e. the `BEGINRELEASENOTES ... ENDRELEASENOTES`).
>Yes, I edit the first comment block and hope this is correct?

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
```diff
@@ -104,69 +103,72 @@ class IsolatedLeptonFinderProcessor : public marlin::Processor {
...
-  bool _mergeCloseElectrons = false;
+  bool  _useDressedLeptons    = false;
+  bool  _mergeCloseElectrons  = false;
   float _dressPhotonConeAngle = 0;
   float _mergeLeptonConeAngle = 0;
 
   /** If set to true, uses Pandora particle IDs */
   bool _usePandoraIDs = false;
+
+  /* DEBUG: Map from copied PFO to original PFO */
```
tmadlener: Could you also add **why** you need this map in the comment?

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -177,10 +169,12 @@ IsolatedLeptonFinderProcessor::IsolatedLeptonFinderProcessor() : Processor("Isol
 
 void IsolatedLeptonFinderProcessor::init() {
   streamlog_out(DEBUG) << "   init called  " << std::endl;
+  streamlog_out(ERROR) << "### USING FIXED IsolatedLeptonFinderProcessor (local build) ###" << std::endl;
```
tmadlener:
```suggestion
```
I suppose you mainly have this here for checking whether you run your version.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -469,27 +460,25 @@ bool IsolatedLeptonFinderProcessor::IsIsolatedPolynomial(ReconstructedParticle*
 }
 
 bool IsolatedLeptonFinderProcessor::IsIsolatedJet(ReconstructedParticle* pfo) {
-  // jet-based isolated lepton (LAL algorithm)
+  ReconstructedParticle* orig = _copy2orig.count(pfo) ? _copy2orig[pfo] : pfo;
```
tmadlener: This is doing the lookup in the map twice if the `pfo` is stored in there. I would put this into a dedicated small helper function (since it is used again below), something like

```cpp
/// Try to find the original particle from which this pfo has been created or return it if no original is found
ReconstructedParticle* findOriginal(ReconstructedParticle* pfo) {
  if (const auto it = _copy2orig.find(pfo); it != _copy2orig.end()) {
    return it->second;
  }
  return pfo;
}
```

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -235,26 +229,24 @@ void IsolatedLeptonFinderProcessor::processEvent(LCEvent* evt) {
 
   // process and save leptons
   for (unsigned int i = 0; i < goodLeptonIndices.size(); ++i) {
-    ReconstructedParticle* pfo_tmp =
-        static_cast<ReconstructedParticle*>(_pfoCol->getElementAt(goodLeptonIndices.at(i)));
-    ReconstructedParticleImpl* pfo = CopyReconstructedParticle(pfo_tmp);
+    ReconstructedParticle*     pfo_tmp = static_cast<ReconstructedParticle*>(_pfoCol->getElementAt(goodLeptonIndices.at(i)));
+    ReconstructedParticleImpl* pfo     = CopyReconstructedParticle(pfo_tmp);
```
tmadlener: These seem to be leaked(?). I can't find where they are either cleaned up again, or added to a collection (which would then assume ownership)

tmadlener (commented):
It looks like you have successfully picked only your changes here. There still seems to be quite a bit of additional formatting going on. In which (software) environment / setup are you running `clang-format`? In the end this doesn't matter too much for now, we can always fix it in the end, when all the other things are done.

I have a few smaller comments / requests below.

On the bigger picture it's not entirely clear to me why this processor creates copies of all PFOs. It also looks to me like they are (at least partially) leaked, but that seems to be pre-existing behavior. Did you by chance run into any memory issues when running this processor on larger samples?

yyzhang666:
> It looks like you have successfully picked only your changes here. There still seems to be quite a bit of additional formatting going on. In which (software) environment / setup are you running `clang-format`? In the end this doesn't matter too much for now, we can always fix it in the end, when all the other things are done.

It's what the ZHH repository are using: /cvmfs/sw.hsf.org/key4hep/releases/2024-10-03/x86_64-almalinux9-gcc14.2.0-opt/llvm/17.0.6-7m4ago/bin/clang-format

> On the bigger picture it's not entirely clear to me why this processor creates copies of all PFOs. It also looks to me like they are (at least partially) leaked, but that seems to be pre-existing behavior. Did you by chance run into any memory issues when running this processor on larger samples?

I only run 12500 events once at most so far, looks normal. They basically use an additional collection to dress leptons and output a collection without isolated leptons.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
yyzhang666: Yes, this looks better, should I also use this function in the getConeEnergy()?

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
yyzhang666: Looks it should be in the line 270
` // Add PFOs to new collections
  evt->addCollection(outPFOsRemovedIsoLepCol.release(), _outputPFOsRemovedIsoLepCollection.c_str());
  evt->addCollection(outIsoLepCol.release(), _outputIsoLepCollection.c_str());
`
Is that mean the ownership transfer?

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
yyzhang666: Yes, I will delete it later

tmadlener:
> It's what the ZHH repository are using: `/cvmfs/sw.hsf.org/key4hep/releases/2024-10-03/x86_64-almalinux9-gcc14.2.0-opt/llvm/17.0.6-7m4ago/bin/clang-format`

Ah, OK. Yeah that then explains it. The version that is used by the `pre-commit` workflow is 20, so there are some differences. In this case, I will fix the formatting at the end, once we have dealt with the rest. :)

Analysis/IsolatedLeptonFinder/include/IsolatedLeptonFinderProcessor.h
```diff
@@ -104,69 +103,86 @@ class IsolatedLeptonFinderProcessor : public marlin::Processor {
...
+ * However, isolation quantities such as the cone energy are defined with
+ * respect to the original input PFO collection.
+ *
+ * Since the copied PFOs are distinct objects, pointer comparisons would
+ * otherwise fail (e.g. when excluding the lepton itself from the cone),
+ * leading to incorrect self-counting in the isolation energy.
+ *
+ * This map is therefore used to recover the association to the original
+ * PFO when computing isolation-related quantities.
+ */
```
tmadlener: very nice. Thank you.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
tmadlener: Yes, please. (it seems to be the exact same logic there, if I am not mistaken).

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
tmadlener: Ah you are right. I missed that all of them are added to a collection. It happens here:

https://github.com/yyzhang666/MarlinReco/blob/538b5f5633bbe854c2460fba221e56d31c41b70e/Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc#L252-L256

and then indeed as you say these collections are added to the event.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
yyzhang666: Yes, I test it with this function and the output is same as before.

tmadlener (commented):
I will push a commit with fixed formatting (which should also make the diff slightly smaller). I will have another look after that, but I think this is in pretty good shape now.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -469,27 +479,26 @@ bool IsolatedLeptonFinderProcessor::IsIsolatedPolynomial(ReconstructedParticle*
 }
 
 bool IsolatedLeptonFinderProcessor::IsIsolatedJet(ReconstructedParticle* pfo) {
-  // jet-based isolated lepton (LAL algorithm)
```
tmadlener, line 472: This comment should probably stay. Unless, it is / was wrong to begin with.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -469,27 +479,26 @@ bool IsolatedLeptonFinderProcessor::IsIsolatedPolynomial(ReconstructedParticle*
...
+  ReconstructedParticle* orig = findOriginal(pfo);
 
-  if (_rpJetMap.find(pfo) == _rpJetMap.end()) {
+  if (_rpJetMap.find(orig) == _rpJetMap.end()) {
     // this is often the case when jet finding fails e.g. due to too few particles in event
+    // Use original address
```
tmadlener: I would move this comment further up. Alternatively, remove it entirely as it is pretty obvious from the code what we are doing, but the **why** we are doing it is missing. In this case the why is probably simply that the `_rpJetMap` stores the original PFOs, so also that is not extremely interesting, resp. can also be derived from the code in a rather straight forward manner.

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -346,14 +357,18 @@ ReconstructedParticleImpl* IsolatedLeptonFinderProcessor::CopyReconstructedParti
   pfo->setParticleIDUsed(pfo_orig->getParticleIDUsed());
   pfo->setGoodnessOfPID(pfo_orig->getGoodnessOfPID());
   pfo->setStartVertex(pfo_orig->getStartVertex());
-  for (unsigned int i = 0; i < pfo->getTracks().size(); i++) {
+
+  // FIXED
```
tmadlener:
```suggestion
```
Looks like another leftover from development.

yyzhang666: Yes, those comments was written earlier and forgot to arrange them

tmadlener (approved):

Analysis/IsolatedLeptonFinder/src/IsolatedLeptonFinderProcessor.cc
```diff
@@ -469,27 +478,26 @@ bool IsolatedLeptonFinderProcessor::IsIsolatedPolynomial(ReconstructedParticle*
 }
 
 bool IsolatedLeptonFinderProcessor::IsIsolatedJet(ReconstructedParticle* pfo) {
-  // jet-based isolated lepton (LAL algorithm)
+    // jet-based isolated lepton (LAL algorithm)
```
tmadlener:
```suggestion
  // jet-based isolated lepton (LAL algorithm)
```

PR #156, opened 2025-12-23
Make sure to only access confusion matrix if it's available

tmadlener:
Without this setting the ConfusionMatrixFileName to an empty value will lead to a segfault because the _confusion_matrix will not be created

BEGINRELEASENOTES
- Fix potential segfault in `WWCategorisationProcessor` if `ConfusionMatrixFileName` is set to an empty string

ENDRELEASENOTES


tmadlener:
Confirmed that the original failure does no longer happen with this fix in place.

PR #158, opened 2026-03-24
Fix crash if unable to calculate theta from the triplet

### LINKED ISSUE #157: TPCDigiProcessor should not crash the entire reconstruction chain.
### Issue

Current implementation of TPCDigiProcessor crashes the entire reconstruction chain if it encounters a `nan`:

https://github.com/iLCSoft/MarlinReco/blob/130f5231043544a2c4d1f3c2315d0216c5376494/TrackDigi/TPCDigi/src/TPCDigiProcessor.cc#L1037-L1042

https://github.com/iLCSoft/MarlinReco/blob/130f5231043544a2c4d1f3c2315d0216c5376494/TrackDigi/TPCDigi/src/TPCDigiProcessor.cc#L1199-L1205

https://github.com/iLCSoft/MarlinReco/blob/130f5231043544a2c4d1f3c2315d0216c5376494/TrackDigi/TPCDigi/src/TPCDigiProcessor.cc#L1346-L1352

https://github.com/iLCSoft/MarlinReco/blob/130f5231043544a2c4d1f3c2315d0216c5376494/TrackDigi/TPCDigi/src/TPCDigiProcessor.cc#L1411-L1416

### Desired

Should not crash the entire reco chain.

dudarboh:
BEGINRELEASENOTES

- Fix #157 

ENDRELEASENOTES


TrackDigi/TPCDigi/src/TPCDigiProcessor.cc
```diff
@@ -1382,37 +1378,22 @@ double TPCDigiProcessor::getPadTheta(CLHEP::Hep3Vector* firstPoint, CLHEP::Hep3V
...
 
-  padTheta = atan((fabs(pathlength1 + pathlength2)) / (fabs(lastPoint->z() - firstPoint->z())));
-
-  // check that the value returned is reasonable
   if (std::isnan(padTheta) || std::isinf(padTheta)) {
-    std::stringstream errorMsg;
-    errorMsg << "\nProcessor: TPCDigiProcessor \n"
-             << "padTheta = " << padTheta << "\n";
-    throw Exception(errorMsg.str());
+    return twopi / 4.0; // return fallback value
```
tmadlener: Same here.

TrackDigi/TPCDigi/src/TPCDigiProcessor.cc
```diff
@@ -1343,12 +1343,8 @@ double TPCDigiProcessor::getPadPhi(CLHEP::Hep3Vector* thisPoint, CLHEP::Hep3Vect
 
   double padPhi = fabs(pointPhi - localPhi);
 
-  // check that the value returned is reasonable
   if (std::isnan(padPhi) || std::isinf(padPhi)) {
-    std::stringstream errorMsg;
-    errorMsg << "\nProcessor: TPCDigiProcessor \n"
-             << "padPhi = " << padPhi << "\n";
-    throw Exception(errorMsg.str());
+    return twopi / 4.0; // return fallback value
```
tmadlener: Since we are in a process here, could we log this case? Not sure about the level. Either its a warning or debug, IMO.

tmadlener (commented):
Given that we are turning a hard error into something that is handled my main question is whether (and how) we should log whenever this happens.

dudarboh:
I have added the warning logging. I think the code should never really reach here after adding the check for ` dr23 / D < 1.0`

PR #159, opened 2026-06-08
Fix TOF seg. fault if a track has no subtracks.

dudarboh:
BEGINRELEASENOTES

- Fix TOFEstimattor processor crash, if Track object with no subtracks is encountered.

ENDRELEASENOTES

Thanks @Victor-Schwan for spotting.

For the ILD reconstruction at ILC, the final track objects (in `MarlinTrkTracks`) were always a combination of Clupartra and SiTracks, which always produced at least one subTrack, even if it pointed to itself.

For the ILD reconstruction at FCCee, one mode of operation is to run Clupatra tracking with the option to pick up silicon hits. This is the most basic approach to achieve track reconstruction for the silicon detectors and the TPC together. In this case, Clupatra tracks don't have any subtracks, so the TOF processor was crashing.


dudarboh:
@gaede

TimeOfFlight/src/TOFUtils.cc
```diff
@@ -62,12 +69,11 @@ const EVENT::TrackState* TOFUtils::getTrackStateAtCalorimeter(EVENT::Track* trac
     }
   }
 
-  // Take the trackState at the calorimeter surface always from the latest curl
-  // Track has only one curl
+  // Track has only one subtrack. Taking the track state of the track itself.
   if (indexOfFirstTPCCurl == nSubTracks - 1)
```
jmcarcell, line 66: Given that the loop above doesn't run when `nSubTracks == 0`?. What about checking here if `nSubTracks == 0`? Then it's a one line change instead of having two checks that return the same.

jmcarcell, line 66:
```suggestion
  if (nSubTracks > 0 && indexOfFirstTPCCurl == nSubTracks - 1)
```

dudarboh: I think then it should be
```
  if (nSubTracks == 0 || indexOfFirstTPCCurl == nSubTracks - 1)
```
no?

This should also work, but I felt like keeping them separate is more readable than this long if condition...

However, now looking at the code, I think if I initialize `indexOfFirstTPCCurl = -1`, then I don't need any extra condition at all and everything remains the same.

jmcarcell: Sorry yes I meant that. Ok, let's merge as it is; failing as soon is possible is good.

jmcarcell (approved):

dudarboh:
@jmcarcell , ah sorry, I read your comment too late, I have already simplified the code to have only one if condition but start the counter at `-1`. I think it is a bit cleaner change.

jmcarcell:
Slightly obscure but OK, if a track is found nothing changes, if not found only when there are no tracks changes which branch it goes.

